PR #43, opened 2020-08-16
Different screen for mobile/fixed session in Step 2

mturnau:


app/src/main/java/io/lunarlogic/aircasting/screens/new_session/connect_airbeam/TurnOnAirBeamViewMvcImpl.kt
```diff
@@ -5,12 +5,14 @@ import android.view.ViewGroup
...
 import io.lunarlogic.aircasting.screens.common.BaseObservableViewMvc
+import io.lunarlogic.aircasting.sensor.Session
 
 class TurnOnAirBeamViewMvcImpl : BaseObservableViewMvc<TurnOnAirBeamViewMvc.Listener>, TurnOnAirBeamViewMvc {
 
     constructor(
-        inflater: LayoutInflater, parent: ViewGroup?): super() {
-        this.rootView = inflater.inflate(R.layout.fragment_turn_on_airbeam, parent, false)
+        inflater: LayoutInflater, parent: ViewGroup?, sessionType: Session.Type): super() {
+        val fragment = getFragmentId(sessionType)
```
lesniakania, line 14: I'd name it `layoutId`. I think we have similar approaches already, so it would be more consistent.

PR #45, opened 2020-08-18
Connected screen, Session details screen.

mturnau:


app/src/main/java/io/lunarlogic/aircasting/screens/new_session/NewSessionWizardNavigator.kt
```diff
@@ -139,7 +139,7 @@ class NewSessionWizardNavigator(
             // for first step
             fragmentTransaction.replace(container, fragment)
         } else {
-            fragmentTransaction.add(container, fragment)
+            fragmentTransaction.replace(container, fragment)
```
mturnau, line 142: CAUTION: I'm not sure if this is ok. I changed it because I noticed by accidence that fragments are stacked on top of each other and I didn't think it was good (a scroll fragment after non scroll one revealed something underneath). At first glance this change doesn't break anything.

lesniakania: I did it intentionally, to be able to use back button to navigate back properly. I will play with this more - thanks for the heads up!

PR #53, opened 2020-09-02
Sign in (designs)

mturnau:


.idea/codeStyles/Project.xml
```diff
@@ -1,6 +1,22 @@
...
+      <option name="PACKAGES_IMPORT_LAYOUT">
+        <value>
+          <package name="" alias="false" withSubpackages="true" />
+          <package name="java" alias="false" withSubpackages="true" />
+          <package name="javax" alias="false" withSubpackages="true" />
+          <package name="kotlin" alias="false" withSubpackages="true" />
+          <package name="" alias="true" withSubpackages="true" />
+        </value>
+      </option>
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
```
mturnau, line 20: no idea what is that, my as added this

app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginController.kt
```diff
@@ -26,13 +27,15 @@ class LoginController(
         mViewMvc.unregisterListener(this)
     }
 
-    override fun onLoginClicked(username: String, password: String) {
+    override fun onLoginClicked(username: String, password: String, usernameInputLayout: TextInputLayout, passwordInputLayout: TextInputLayout) {
         val successCallback = {
```
mturnau, line 31: hm hm doesn't seem too nice way to do this but didn;t have idea where to put that logic

app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginController.kt
```diff
@@ -26,13 +27,15 @@ class LoginController(
...
             performSessionSync()
             MainActivity.start(mContext)
         }
         val message = mContext.getString(R.string.invalid_credentials_message)
         val errorCallback = {
+            usernameInputLayout.error = " "
+            passwordInputLayout.error = " "
             val toast = Toast.makeText(mContext, message, Toast.LENGTH_LONG)
```
mturnau, line 39: this is a workaround to show a red border on error but no text

app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginController.kt
lesniakania: I think it should belong to the view. I will call you and we can talk about this, it will be easier than explaining it here I guess.

PR #59, opened 2020-10-05
Changed tags on session cards to session type and instrument name.

mturnau:


app/src/androidTest/java/io/lunarlogic/aircasting/FixedSessionTest.kt
```diff
@@ -140,7 +140,7 @@ class FixedSessionTest {
         Thread.sleep(4000)
 
         onView(withId(R.id.session_name)).check(matches(withText("Ania's fixed outdoor session")))
-        onView(withId(R.id.session_tags)).check(matches(withText("tag1, tag2")));
+        onView(withId(R.id.session_info)).check(matches(withText("Fixed, AirBeam2")));
```
mturnau, line 143: Had a problem with naming. Emm.

app/src/main/java/io/lunarlogic/aircasting/database/DatabaseProvider.kt
```diff
@@ -48,7 +48,7 @@ class DatabaseProvider {
                 mAppDatabase = Room.databaseBuilder(
                     mContext,
                     AppDatabase::class.java, DB_NAME
-                ).build()
+                ).fallbackToDestructiveMigration().build()
```
mturnau, line 51: I had an error "a migration from 16 to 17 needed" and this was supposed to be solution (SO) but maybe I should do something else.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -116,7 +115,7 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
 
         mDateTextView.text = session?.durationString()
         mNameTextView.text = session?.name
-        mTagsTextView.text = session?.tagsString()
+        mInfoTextView.text = session?.displayedType?.capitalize() + ", " + session?.instrumentNameString()
```
mturnau, line 118: I could move this somewhere but 1) view impelementation looked like the best place to do it 2) I reckon putting it in a separate method may be actually making it harder to read. But I'm open for suggestions

app/src/main/java/io/lunarlogic/aircasting/sensor/MeasurementStream.kt
```diff
@@ -78,6 +78,8 @@ class MeasurementStream(
 
     val detailedType: String?
 
+    val sensorShortName: String?
```
mturnau, line 81: naming problems continue

app/src/main/java/io/lunarlogic/aircasting/sensor/Session.kt
```diff
@@ -175,6 +175,10 @@ class Session(
         return tags.joinToString(", ")
     }
 
+    fun instrumentNameString(): String? {
```
mturnau, line 178: and again... Michael wrote instrument type so I reckoned it's good but isn't it sensor name? device name?? OMG

app/src/main/java/io/lunarlogic/aircasting/database/DatabaseProvider.kt
lesniakania: Yeah, you can do this temporarily, but we should start adding migrations if we change anything in the schema... but it's my fault - I didn't add one last time I changed schema... But we should not commit this, I think

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
lesniakania: Yeah, it's fine. You could use interpolation though `${session?.displayedType?.capitalize()}, ${session?.instrumentNameString()}`.

app/src/androidTest/java/io/lunarlogic/aircasting/FixedSessionTest.kt
```diff
@@ -197,6 +197,6 @@ class FixedSessionTest {
         Thread.sleep(4000)
 
         onView(withId(R.id.session_name)).check(matches(withText("Ania's fixed indoor session")))
-        onView(withId(R.id.session_tags)).check(matches(withText("tag1, tag2")));
+        onView(withId(R.id.session_info)).check(matches(withText("Fixed: ")));
```
mturnau, line 200: @lesniakania For some reason sensor package name doesn't work here for Fixed session but it does work wof Mobile! I couldn't quickly decipher how it works in those tests so maybe you can help me (or we can elave it like this).

lesniakania: I think it's because for mobile sessions we are getting measurements from the device and we save them as they coming. But for fixed sessions, the device is communicating itself directly to the backend and our app has an additional thread to download measurements from the server. And since we have all the API calls mocked with `MockWebServer` it's not working here. I guess we should just leave it like that now...

lesniakania (approved):

PR #65, opened 2020-10-19
Charts

mturnau:


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ChartAveragesCreator.kt
```diff
@@ -0,0 +1,149 @@
+package io.lunarlogic.aircasting.screens.dashboard
+
+import com.github.mikephil.charting.data.Entry
+import com.google.common.collect.Lists
+import io.lunarlogic.aircasting.sensor.Measurement
+import io.lunarlogic.aircasting.sensor.MeasurementStream
+import java.util.*
+import java.util.concurrent.CopyOnWriteArrayList
+
+class ChartAveragesCreator {
```
mturnau: When I thought about how I'd like to refactor this, it made sense to me to maybe divide it into two classes (mobile and fixed) and make the code more readable. However, I don't think it is absolutely necessary right now because it would take some time. So unless you have some quick ideas on how to improve this, I'd leave it like this for now and take care of it in the future.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -75,6 +101,7 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
 
     protected abstract fun showMeasurementsTableValues(): Boolean
     protected abstract fun buildBottomSheet(): BottomSheet?
+    protected abstract fun chartUnitLabel(): Int
```
mturnau, line 105: I had a feeling overloading method is not the best solution in this case but I didn't want to use an "if" either.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/following/FollowingSessionViewMvcImpl.kt
```diff
@@ -69,4 +69,8 @@ class FollowingSessionViewMvcImpl:
 
         mExpandSessionButton.visibility = View.VISIBLE
     }
+
+    override fun chartUnitLabel(): Int {
+        return R.string.fixed_session_units_label
```
mturnau, line 74: emm not sure if that's right

PR #72, opened 2020-11-09
Simplify card (mobile dormant)

mturnau:


app/src/main/java/io/lunarlogic/aircasting/screens/common/MeasurementsTableContainer.kt
```diff
@@ -86,6 +89,7 @@ class MeasurementsTableContainer {
     ) {
         mSessionPresenter = sessionPresenter
         mOnMeasurementStreamChanged = onMeasurementStreamChanged
+        mDisplayAvarages = mSessionPresenter?.isMobileDormant() ?: false
```
mturnau, line 92: Not sure if this object should contain the logic behind it. On the other hand, we do pass sessionPresenter here so the info is there.

app/src/main/java/io/lunarlogic/aircasting/screens/map/MapViewMvcImpl.kt
```diff
@@ -90,7 +92,11 @@ class MapViewMvcImpl: BaseObservableViewMvc<MapViewMvc.Listener>, MapViewMvc, Ma
 
         mMapContainer.bindSession(mSessionPresenter)
         mMeasurementsTableContainer.bindSession(mSessionPresenter, this::onMeasurementStreamChanged)
-        mStatisticsContainer.bindSession(mSessionPresenter)
+        if(mSessionPresenter?.isMobileDormant() ?: false) {
```
mturnau, line 95: I was thinking maybe we should have separate MapViewMvcImpl for each session type like we have with sessionView? to avoid those ifs

app/src/main/java/io/lunarlogic/aircasting/screens/map/MapViewMvcImpl.kt
```diff
@@ -110,12 +116,13 @@ class MapViewMvcImpl: BaseObservableViewMvc<MapViewMvc.Listener>, MapViewMvc, Ma
         mSessionDateTextView?.text = session.durationString()
         mSessionNameTextView?.text = session.name
         mSessionTagsTextView?.text = session.infoString()
+        if (mSessionPresenter?.isMobileDormant() ?: false) mSessionMeasurementsDescription?.text = context.getString(R.string.session_avg_measurements_description)
```
mturnau, line 119: same as above

app/src/main/java/io/lunarlogic/aircasting/screens/common/MeasurementsTableContainer.kt
```diff
@@ -182,18 +186,18 @@ class MeasurementsTableContainer {
         } catch(e: IndexOutOfBoundsException) {}
     }
 
-    private fun bindLastMeasurement(stream: MeasurementStream) {
-        val measurement = stream.measurements.lastOrNull() ?: return
+    private fun bindMeasurement(stream: MeasurementStream) {
+        val measurementValue = (if (mDisplayAvarages) stream.getAvgMeasurement() else stream.measurements.lastOrNull()?.value) ?: return
```
lesniakania, line 190: I'd split this line because it's a bit hard to read it

app/src/main/java/io/lunarlogic/aircasting/screens/map/MapViewMvcImpl.kt
lesniakania: Yeah, seems like a good idea... Maybe you can try to timebox it to an hour or two and if it occurs a bigger task, we can do it in the future maybe - what do you think?

And if we would leave these ifs, I'd change them too `if (mSessionPresenter?.isMobileDormant() == true)` - I think it would be more consistent with other places and more readable.

app/src/main/java/io/lunarlogic/aircasting/screens/map/MapActivity.kt
```diff
@@ -11,21 +11,32 @@ import io.lunarlogic.aircasting.exceptions.ErrorHandler
...
+        val SESSION_UUID_KEY = "SESSION_UUID"
+        val SESSION_TYPE_KEY = "SESSION_TYPE"
+        val SESSION_STATUS_KEY = "SESSION_STATUS"
 
-        fun start(context: Context?, sessionUUID: String, sensorName: String?) {
+        fun start(
+            context: Context?,
+            sensorName: String?,
+            sessionUUID: String,
+            sessionType: Int,
```
lesniakania, line 31: I'd pass `Session.Type` and `Session.Status` here and get `.value` while putting to the intent, but it's me being picky ;]

app/src/main/java/io/lunarlogic/aircasting/screens/map/MapViewMvcImplFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+            supportFragmentManager: FragmentManager?,
+            sessionType: Session.Type,
+            sessionStatus: Session.Status
+        ): MapViewMvcImpl {
+            return when {
+                sessionType == Session.Type.MOBILE && sessionStatus == Session.Status.FINISHED -> MapViewMobileDormantMvcImpl(inflater, parent, supportFragmentManager)
+                sessionType == Session.Type.MOBILE && sessionStatus != Session.Status.FINISHED -> MapViewMobileActiveMvcImpl(inflater, parent, supportFragmentManager)
+                sessionType == Session.Type.FIXED -> MapViewFixedMvcImpl(inflater, parent, supportFragmentManager)
+                else -> MapViewFixedMvcImpl(inflater, parent, supportFragmentManager)
+            }
```
lesniakania, line 23: I'd go for something like that

```kotlin
if (sessionType == Session.Type.FIXED) {
  return MapViewFixedMvcImpl(inflater, parent, supportFragmentManager)
}

return when(sessionStatus) {
  Session.Status.FINISHED -> MapViewMobileDormantMvcImpl(inflater, parent, supportFragmentManager)
  Session.Status.RECORDING -> MapViewMobileActiveMvcImpl(inflater, parent, supportFragmentManager)
  else -> MapViewFixedMvcImpl(inflater, parent, supportFragmentManager)
}
```

or extract when conditions from your code, to make it more readable

app/src/main/java/io/lunarlogic/aircasting/sensor/MeasurementStream.kt
```diff
@@ -124,6 +124,15 @@ class MeasurementStream(
...
     }
 
+    fun getAvgMeasurement(): Double? {
+        if(measurements.isEmpty()) return null
+        return calculateSum() / measurements.size
+    }
+
+    private fun calculateSum(): Double {
+        return measurements.sumByDouble { it.value }
+    }
```
lesniakania, line 134: there is `calculateSum` method in the `StatisticsContainer` which does the same - maybe we can now use this method there?

app/src/main/java/io/lunarlogic/aircasting/sensor/Session.kt
```diff
@@ -72,13 +72,21 @@ class Session(
 
     enum class Type(val value: Int){
         MOBILE(0),
-        FIXED(1)
+        FIXED(1);
+
+        companion object {
+            fun fromInt(value: Int) = Type.values().first { it.value == value }
```
lesniakania, line 78: There is `SessionTypeConverter` which does the same - we can now use this method there.

app/src/main/java/io/lunarlogic/aircasting/sensor/Session.kt
```diff
@@ -72,13 +72,21 @@ class Session(
...
     }
 
     enum class Status(val value: Int){
         NEW(-1),
         RECORDING(0),
-        FINISHED(1)
+        FINISHED(1);
+
+        companion object {
+            fun fromInt(value: Int) = Status.values().first { it.value == value }
```
lesniakania, line 88: There is `SessionStatusConverter` which does the same - we can now use this method there.

app/src/main/java/io/lunarlogic/aircasting/sensor/MeasurementStream.kt
mturnau: good point, actually my intention was to move it to stream and use in statistics.

PR #74, opened 2020-11-16
Chart refreshing smarter

mturnau:


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/charts/Chart.kt
```diff
@@ -42,20 +43,26 @@ class Chart {
...
-            drawChart()
-            setTimesAndUnit()
+        if(mChartRefreshService.shouldBeRefreshed()) {
+            val session = sessionPresenter?.session
+            mSessionPresenter = sessionPresenter
+            mEntries =
+                sessionPresenter?.chartData?.getEntries(sessionPresenter.selectedStream) ?: listOf()
+
+            resetChart()
+            if (session != null && session?.streams.count() > 0) {
```
lesniakania, line 60: Do you need `session != null` here if you have  `session?.` later already?

PR #78, opened 2020-11-27
Updating fixed endTime

mturnau:


app/src/main/java/io/lunarlogic/aircasting/models/MeasurementStream.kt
```diff
@@ -119,6 +119,7 @@ class MeasurementStream(
 
         if (amount >= allMeasurements.size) return allMeasurements
 
+        // TODO: seems like a duplication
```
mturnau, line 122: I promise I'll fix it soon

app/src/main/java/io/lunarlogic/aircasting/networking/services/DownloadMeasurementsCallback.kt
```diff
@@ -30,8 +32,11 @@ class DownloadMeasurementsCallback(
     ) {
         if (response.isSuccessful) {
             val body = response.body()
+
             body?.streams?.let { streams ->
                 DatabaseProvider.runQuery {
+                    session.endTime = DateConverter.fromString(body?.end_time)
```
mturnau, line 38: not sure if this is a place and means to do it but I reckon this is roughly the place where we ensure right things are saved to DB? please correct me if you have a better idea.

lesniakania: Yeah, I think it's good place, but I'd move this line to the other method (e.g. `updateSessionEndTime`) and combine with the actual update to db. I'm actually not sure, why this update is in the saving measurements loop, I think it should not be there, because it's not updating anything measurements wise in the session table.

So basically you'd have:
```kotlin
fun updateSessionEndTime(endTime: Date) {
  session.endTime = DateConverter.fromString()
  sessionsRepository.update(session)
}
```

PR #79, opened 2020-11-30
Settings - Sign out

michalrazny2:
I think that sign out button should work now.

My doubts we didn't discuss yet:
- I overriden onBackPressed() in LoginActivity to prevent pressing back button after Sign out,
- Probably this version will require some additional changes in layouts,
- For now 'my account' package (AccountActivity etc.) is inside 'screens.settings' package, not sure if it should be organised this way


.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="MarkdownUnresolvedFileReference" enabled="true" level="WARNING" enabled_by_default="true" />
+  </profile>
+</component>
```
lesniakania, line 6: We probably should do something about this. I guess we should all update Android Studio or something?
I'm not sure. But it looks like this is something we should not commit changes back and forth.
I was ignoring this because of the lack of time, but you can check it out maybe....?

app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginActivity.kt
```diff
@@ -45,4 +45,8 @@ class LoginActivity: AppCompatActivity() {
         super.onStop()
         controller!!.onStop()
     }
+
+    override fun onBackPressed() {
+
```
lesniakania, line 50: You add comment here why it's needed. E.g "Doing nothing just to prevent from going back after the logout"

app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginController.kt
```diff
@@ -9,7 +9,7 @@ import io.lunarlogic.aircasting.screens.create_account.CreateAccountActivity
 import io.lunarlogic.aircasting.screens.main.MainActivity
 import io.lunarlogic.aircasting.screens.login.LoginService
 
-class LoginController(
+class  LoginController(
```
lesniakania, line 12: Neat picking, but I guess it's not intended change ; ]?

app/src/main/java/io/lunarlogic/aircasting/screens/settings/SettingsController.kt
```diff
@@ -0,0 +1,29 @@
...
+import android.view.View
+import io.lunarlogic.aircasting.screens.settings.myaccount.MyAccountActivity
+
+class SettingsController(
+    private val mContext: Context?,
+    private val mViewMvc: SettingsViewMvc
+) : SettingsViewMvc.Listener {
+
+    fun onCreate(){
+        // Todo: do i really need this one??
```
lesniakania, line 14: You don't ; ) If you don't need to do anything here, just remove the method.

app/src/main/java/io/lunarlogic/aircasting/screens/settings/SettingsViewMvcImpl.kt
```diff
@@ -0,0 +1,34 @@
...
+
+        val myAccountButton = rootView?.findViewById<Button>(R.id.myAccount_Button)
+        myAccountButton?.setOnClickListener {
+            onMyaccountClicked()
+        }
+
+    }
+
+    private fun onMyaccountClicked() {
+        //Todo: going to MyAccountActivity
```
lesniakania, line 26: You can remove this comment already, I guess?

app/src/main/java/io/lunarlogic/aircasting/screens/settings/myaccount/MyAccountActivity.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        (application as AircastingApplication)
+            .appComponent.inject(this)
+
+        val view = MyAccountViewMvcImpl(layoutInflater, null)
+        controller = MyAccountController(this, view, settings)
+
+        setContentView(view.rootView)
+        AppBar.setup(view.rootView, this)
+
```
lesniakania, line 40: Neat picking again - I'd remove this extra space ; ]

lesniakania (approved):

lesniakania:
Awesome job 👏 You can also try to add tests : > It may be more important just for the sake you getting familiar with them : ]

.idea/inspectionProfiles/Project_Default.xml
michalrazny2: Hmm... my version of Android Studio is 4.1.1. Is this the reason of change here? I can check it in a second

lesniakania: I'm not sure, but I think it might be. My is 4.0.1, so actually, I should update then ; p @mturnau what's yours?

PR #80, opened 2020-12-01
Fix now measurements - V1

mturnau:
This is a first version of fixing "Now" measurements in statistics container. This one is using an event on database update. it is not ideal because DB update != sessionPresenter update, so we have difference between measurements table and statistics table sometimes.


mturnau:
I decided to use different approach (PR #81)

PR #82, opened 2020-12-02
Graph view mobile dormant - remove statistics container.

mturnau:
https://trello.com/c/mxfPMLtx/1009-mobile-dormant-graph-remove-the-box-that-displays-the-avg-now-and-peak


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/graph/GraphViewFixedMvcImpl.kt
```diff
@@ -4,7 +4,7 @@ import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.fragment.app.FragmentManager
 
-class GraphViewFixedMvcImpl(
+open class GraphViewFixedMvcImpl(
```
lesniakania, line 7: I think it would be better to use `abstract` here as you have in `GraphViewMobileMvcImpl `

PR #84, opened 2020-12-07
Do not load measurements for dormant and fixed sessions

lesniakania:
Loading measurements for dormant and fixed sessions results in too much data loaded at once.
Fixes https://trello.com/c/SfeWRrsf/925-fixed-mobile-dormant-after-having-synced-fixed-mobile-dormant-sessions-they-should-always-be-present


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -17,99 +12,24 @@ import io.lunarlogic.aircasting.networking.services.DownloadMeasurementsService
...
-import kotlinx.coroutines.CoroutineScope
 
 
 abstract class SessionsController(
     private val mRootActivity: FragmentActivity?,
     private val mViewMvc: SessionsViewMvc,
     private val mSessionsViewModel: SessionsViewModel,
-    private val mLifecycleOwner: LifecycleOwner,
-    private val mSettings: Settings
+    mLifecycleOwner: LifecycleOwner,
```
mturnau, line 23: unused?

lesniakania: yes, thanks for catching this, I'll remove it

PR #85, opened 2020-12-08
Settings mvp

lesniakania:


app/src/androidTest/java/io/lunarlogic/aircasting/MobileSessionTest.kt
```diff
@@ -97,6 +97,7 @@ class MobileSessionTest {
         onView(allOf(withId(R.id.navigation_lets_start), isDisplayed())).perform(click())
 
         onView(withId(R.id.mobile_session_start_card)).perform(click())
+        Thread.sleep(5000)
```
mturnau, line 100: I know sleeps are sometimes necessary for tests but we should always aim for adding the shortest sleeps possible. 5s is a long one, just make sure it's the shortest working sleep.

app/src/main/java/io/lunarlogic/aircasting/lib/Settings.kt
```diff
@@ -2,12 +2,17 @@ package io.lunarlogic.aircasting.lib
...
 import android.app.Application
 import android.content.SharedPreferences
+import android.util.Log
 
 open class Settings(mApplication: Application) {
     private val PRIVATE_MODE = 0
     protected val PREFERENCES_NAME = "preferences"
     protected val EMAIL_KEY = "email"
     protected val AUTH_TOKEN_KEY = "auth_token"
+    protected val CROWD_MAP_SETTING = "crowd_map_setting"
```
mturnau, line 12: no need to call them _SETTING (we know they are settings). The suffix could be _KEY though.

app/src/main/java/io/lunarlogic/aircasting/lib/Settings.kt
```diff
@@ -46,4 +51,34 @@ open class Settings(mApplication: Application) {
         editor.clear()
         editor.apply()
     }
+
+    fun crowdMapSettingSwitched(){
+        val check = sharedPreferences.getBoolean(CROWD_MAP_SETTING, false)
+        val editor = sharedPreferences.edit()
+        editor.putBoolean(CROWD_MAP_SETTING, !check)
```
mturnau, line 58: you could create another method like we already have but with different signature, e.g. 
` open protected fun saveToSettings(key: String, value: Boolean) `
For saving boolean values. Or think of DRYing it out a bit more even (making saveToSettings more universal).

app/src/main/java/io/lunarlogic/aircasting/screens/settings/BackendSettingsDialog.kt
```diff
@@ -0,0 +1,42 @@
...
+    private fun settingsConfirmed() {
+        val adress_value = mView.adress_edittext.text.toString()
+        val port_value = mView.port_edittext.text.toString()
+        mSettings.backendSettingsChanged(adress_value, port_value)
+        Log.i("SETTINGS_DIALOG", "Adress and port values added to sharedPreferences")
+        dismiss()
+    }
+
+
+}
```
mturnau, line 42: Please setup Android Studio to automatically add an newline in the end of the file.
https://thoughtbot.com/blog/no-newline-at-end-of-file

app/src/main/java/io/lunarlogic/aircasting/screens/main/MainActivity.kt
```diff
@@ -46,7 +47,8 @@ class MainActivity: AppCompatActivity() {
         controller?.onCreate()
 
         setContentView(view.rootView)
-        setSupportActionBar(findViewById(R.id.topAppBar))
+        AppBar.setup(view.rootView, this)
```
mturnau, line 50: For some reason this line causes irrational fear in me! I think it has something to do with my bad experiences with AppBar configuration. Just let's make sure it doesn't break the designs anywhere...

app/src/main/java/io/lunarlogic/aircasting/lib/Settings.kt
```diff
@@ -2,12 +2,17 @@ package io.lunarlogic.aircasting.lib
...
 import android.content.SharedPreferences
+import android.util.Log
 
 open class Settings(mApplication: Application) {
     private val PRIVATE_MODE = 0
     protected val PREFERENCES_NAME = "preferences"
     protected val EMAIL_KEY = "email"
     protected val AUTH_TOKEN_KEY = "auth_token"
+    protected val CROWD_MAP_SETTING = "crowd_map_setting"
+    protected val ADRESS_BACKEND_SETTING = "adress_backend_setting"
```
lesniakania, line 13: I'd name it `BACKEND_URL_KEY`

app/src/main/java/io/lunarlogic/aircasting/lib/Settings.kt
```diff
@@ -2,12 +2,17 @@ package io.lunarlogic.aircasting.lib
...
+    protected val PORT_BACKEND_SETTING = "port_backend_setting"
```
lesniakania, line 14: I'd name it `BACKEND_PORT_KEY`

app/src/main/java/io/lunarlogic/aircasting/screens/settings/SettingsViewMvcImpl.kt
```diff
@@ -20,13 +24,34 @@ class SettingsViewMvcImpl : BaseObservableViewMvc<SettingsViewMvc.Listener>, Set
             onMyaccountClicked()
         }
 
+        val contributeToCrowdMapSwitch = rootView?.findViewById<SwitchMaterial>(R.id.crowd_map_settings_switch)
```
mturnau, line 27: if you add this line in the imports: `import kotlinx.android.synthetic.main.fragment_settings.view.*` you can then use `rootView?.crowd_map_settings_switch` (magic!) I know we don't use it everywhere but we can try using it in the new places, it looks clearer I guess.

app/src/main/res/layout/activity_myaccount.xml
```diff
@@ -4,40 +4,40 @@
...
-        android:layout_marginTop="108dp"
-        android:drawableEnd="@drawable/ic_baseline_chevron_right_white"
-        android:text="@string/sign_out"
-        android:textAlignment="viewStart"
-        app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintHorizontal_bias="0.0"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/app_bar_layout" />
+        android:layout_height="260dp"
+        android:layout_marginTop="12dp"
```
mturnau, line 16: we were trying to use predefined values for margins etc, `@keyline_n` where n is multiple of 4

app/src/main/res/layout/activity_myaccount.xml
```diff
@@ -4,40 +4,40 @@
...
-        android:layout_height="60dp"
-        android:layout_marginTop="108dp"
-        android:drawableEnd="@drawable/ic_baseline_chevron_right_white"
-        android:text="@string/sign_out"
-        android:textAlignment="viewStart"
-        app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintHorizontal_bias="0.0"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/app_bar_layout" />
+        android:layout_height="260dp"
```
mturnau, line 15: We aim for not hardcoding heights so it works nicely on different devices. Sometimes it's not easy to make things look good just using constraints but most often you can do it.

app/src/main/res/layout/activity_myaccount.xml
```diff
@@ -4,40 +4,40 @@
...
\ No newline at end of file
+        <TextView
+            android:id="@+id/login_state_textView"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:text="@string/my_account_info"
+            android:textAlignment="viewStart"
+            android:layout_marginTop="25dp"
+            android:layout_marginStart="50dp"
+            android:layout_marginEnd="50dp"/>
```
mturnau, line 28: did you put these margins to position the text view? It was supposed to be centered? I suppose there may be a better way. If you have problems and you want, we can pair on constraint layouts. By all means, I am no expert but I can try ;)

app/src/main/res/values/colors.xml
```diff
@@ -14,6 +14,7 @@
     <color name="aircasting_grey_500">#949599</color>
     <color name="aircasting_grey_600">#858B93</color>
     <color name="aircasting_grey_700">#737578</color>
+    <color name="aircasting_grey_800">#464C55</color>
```
mturnau, line 17: Did you get this color from the designs? At some point, we realized we have too many shades of grey in this app and Pina told us to use grey_700 in all grey texts, so have it in mind.

michalrazny2: Hmmm... I am not sure if this question is still valid- As far as I can see there is no usage of aircasting_grey_800 anywhere in layouts now. I believe that this aircasting_grey_800 was supposed to be <almost black font> which appears on some headers (?).
@lesniakania

PR #95, opened 2020-12-21
Disable mapping- visual changes

michalrazny2:
Very small PR connected with <settings: copy for "Contribute to CrowdMap" & "Disable mapping"> task from Trello.


app/src/main/res/layout/fragment_settings.xml
```diff
@@ -105,7 +105,7 @@
                 android:id="@+id/map_settings_switch"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:checked="true"
+                android:checked="false"
```
lesniakania, line 108: We set `checked` in the `SettingsViewMvcImpl` for both `map_settings_switch` and `crowd_map_settings_switch`, so you can delete it from here maybe because it might be confusing for someone.

app/src/main/res/values/strings.xml
```diff
@@ -192,8 +192,11 @@
     <string name="settings_backend_settings">Backend settings</string>
     <string name="settings_backend_url">Url</string>
     <string name="settings_backend_port">Port</string>
+    <string name="map_enable_description"><![CDATA[Turns off GPS tracking & session syncing. Use \'Share file\' to retrieve your measurements via email.]]></string>
```
lesniakania, line 195: Is this `![CDATA[` part in purpose here? Also, I wonder about the "Share file" feature (we don't have it ;]), but I'll just ask Micheal on the Trello ;]

PR #97, opened 2020-12-21
Fixes for disconnect view

lesniakania:
Fixes:

https://trello.com/c/w1HHV4Lq/1056-v146-fixed-following-after-configuring-an-ab3-to-record-a-fixed-session-the-measurement-device-disconnected-message-is-shown-und
and

https://trello.com/c/a5zpwHjk/1063-mobile-active-disconnect-ab3-wrong-card-displayed

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/MeasurementsTableContainer.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/MeasurementsTableContainer.kt
@@ -27,7 +27,8 @@ class MeasurementsTableContainer {
 
     private var mSelectable: Boolean
     private var mDisplayValues: Boolean
-    private var mDisplayAvarages: Boolean = false
+    private var mDisplayAvarages = false
+    private var mDisplayDisconnectedIndicator = false
 
     private val mMeasurementStreams: MutableList<MeasurementStream> = mutableListOf()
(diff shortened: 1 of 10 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/MeasurementsTableContainer.kt
```diff
@@ -224,7 +232,7 @@ class MeasurementsTableContainer {
         val circleView = valueView.findViewById<ImageView>(R.id.circle_indicator)
         val valueTextView = valueView.findViewById<TextView>(R.id.measurement_value)
 
-        if (mSessionPresenter?.isDisconnected() == true) {
+        if (mDisplayDisconnectedIndicator && mSessionPresenter?.isDisconnected() == true) {
```
mturnau, line 235: hm, can we move `isDisconnected()` condition to setting mDisplayDisconnectedIndicator value? And reset it every time we update session? Not sure if it makes sense..hm...

lesniakania: yeah, actually this might be good idea :)

lesniakania: I ended up removing `mDisplayDisconnectedIndicator` completely ; p

PR #99, opened 2020-12-22
Save disable mapping and contribute settings per session

lesniakania:
https://trello.com/c/bTq5WreJ/1068-disable-mapping-and-contribute-to-crowdmap-settings-should-be-saved-per-session

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
@@ -27,7 +27,10 @@ data class SessionDBObject(
     @ColumnInfo(name = "status") val status: Session.Status = Session.Status.NEW,
     @ColumnInfo(name = "version") val version: Int = 0,
     @ColumnInfo(name = "deleted") val deleted: Boolean = false,
-    @ColumnInfo(name = "followed_at") val followedAt: Date? = null
+    @ColumnInfo(name = "followed_at") val followedAt: Date? = null,
+    @ColumnInfo(name = "contribute") val contribute: Boolean = false,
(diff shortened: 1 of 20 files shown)


app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
```diff
@@ -27,7 +27,10 @@ data class SessionDBObject(
     @ColumnInfo(name = "status") val status: Session.Status = Session.Status.NEW,
     @ColumnInfo(name = "version") val version: Int = 0,
     @ColumnInfo(name = "deleted") val deleted: Boolean = false,
-    @ColumnInfo(name = "followed_at") val followedAt: Date? = null
+    @ColumnInfo(name = "followed_at") val followedAt: Date? = null,
+    @ColumnInfo(name = "contribute") val contribute: Boolean = false,
+    @ColumnInfo(name = "locationless") val locationless: Boolean = false
```
mturnau, line 32: locationless :O hm... send_location? Or sth?

lesniakania: this is the name from the old app ; d It's actually not about sending location, but generally about not sending it to the backend and this session will not have also location, because it means you are disabling GPS completely...
I also didn't like this name while I was looking in the old app code, but I actually can't come up with better... ; p
I will think a bit more and see how it goes ; p

lesniakania: hmmm maybe something related to this setting name, so e.g. `mapping_enabled` or `mapping_disabled`, hm? I will merge it for now, but maybe we can extract this as a separate task/chore, and maybe @michalrazny2 can do this?

PR #104, opened 2021-01-07
Share session new

michalrazny2:
Share session on new version of master

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
@@ -29,7 +29,8 @@ data class SessionDBObject(
     @ColumnInfo(name = "deleted") val deleted: Boolean = false,
     @ColumnInfo(name = "followed_at") val followedAt: Date? = null,
     @ColumnInfo(name = "contribute") val contribute: Boolean = false,
-    @ColumnInfo(name = "locationless") val locationless: Boolean = false
+    @ColumnInfo(name = "locationless") val locationless: Boolean = false,
+    @ColumnInfo(name = "url_location") val urlLocation: String? = null
 
 ) {
     @PrimaryKey(autoGenerate = true)
@@ -52,7 +53,8 @@ data class SessionDBObject(
                 session.deleted,
                 session.followedAt,
                 session.contribute,
-                session.locationless
+                session.locationless,
+                session.urlLocation
             )
 }
 
--- a/app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
@@ -5,6 +5,6 @@ import androidx.sqlite.db.SupportSQLiteDatabase
 
 val MIGRATION_21_22 = object : Migration(21, 22) {
     override fun migrate(database: SupportSQLiteDatabase) {
-        database.execSQL("ALTER TABLE `sessions` ADD `urlLocation` TEXT")
+        database.execSQL("ALTER TABLE `sessions` ADD `urlLocation` INTEGER NOT NULL DEFAULT 0")
     }
 }
--- a/app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
@@ -1,0 +1,10 @@
+package io.lunarlogic.aircasting.database.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val MIGRATION_21_22 = object : Migration(21, 22) {
+    override fun migrate(database: SupportSQLiteDatabase) {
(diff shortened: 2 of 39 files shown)


app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
```diff
@@ -0,0 +1,10 @@
+package io.lunarlogic.aircasting.database.migrations
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+val MIGRATION_21_22 = object : Migration(21, 22) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("ALTER TABLE `sessions` ADD `urlLocation` INTEGER NOT NULL DEFAULT 0")
+    }
```
michalrazny2, line 9: I tried a couple of possibilities under execSQL() but none of them worked correctly

app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
```diff
@@ -29,7 +29,8 @@ data class SessionDBObject(
     @ColumnInfo(name = "deleted") val deleted: Boolean = false,
     @ColumnInfo(name = "followed_at") val followedAt: Date? = null,
     @ColumnInfo(name = "contribute") val contribute: Boolean = false,
-    @ColumnInfo(name = "locationless") val locationless: Boolean = false
+    @ColumnInfo(name = "locationless") val locationless: Boolean = false,
+    @ColumnInfo(name = "url_location") val urlLocation: String? = null
```
lesniakania, line 33: you have `String?` here but `NOT NULL` in schema - these should much - either `String?` and without `NOT NULL` or `NOT NULL` and `String` (not `String?`)

app/src/main/java/io/lunarlogic/aircasting/database/migrations/migration_22.kt
lesniakania: also this should be `TEXT`, probably without `NOT NULL` and `DEFAULT`

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -113,19 +120,53 @@ abstract class SessionsController(
...
+    override fun onShareLinkPressed(session: Session, sensor: String) { // handling button in ShareSessionBottomSheet
+        openShareIntentChooser(session, sensor)
+        shareDialog?.dismiss()
+    }
+
+    override fun onShareFilePressed(session: Session, emailInput: String) { // handling button in ShareSessionBottomSheet
+        val event = ExportSessionEvent(session, emailInput)
+        EventBus.getDefault().post(event)
+
+        shareDialog?.dismiss()
```
lesniakania, line 140: Actually, all these dialog?.dismiss() should happen in the dialogs themselves - before or after they notify their listeners. It should be their responsibility. And you can then remove `shareDialog` and `editDialog` variables from this controller.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ShareSessionBottomSheet.kt
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        val shareFileButton = view?.findViewById<Button>(R.id.share_file_button)
+        shareFileButton?.setOnClickListener {
+            shareFilePressed()
+        }
+
+        val cancelButton = view?.findViewById<Button>(R.id.cancel_button)
+        cancelButton?.setOnClickListener {
+            mListener.onCancelPressed()
```
lesniakania, line 67: here you can just dismiss the dialog, there is no need to notify the listener

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ShareSessionBottomSheet.kt
```diff
@@ -0,0 +1,112 @@
...
+import io.lunarlogic.aircasting.models.Session
+
+class ShareSessionBottomSheet(
+    private val mListener: ShareSessionBottomSheet.Listener,
+    val session: Session
+): BottomSheetDialogFragment() {
+    interface Listener{
+        fun onShareLinkPressed(session: Session, sensor: String)
+        fun onShareFilePressed(session: Session, emailInput: String)
+        fun onCancelPressed()
```
lesniakania, line 24: Having in mind other comments this would not be needed then

app/src/main/java/io/lunarlogic/aircasting/exceptions/Exceptions.kt
```diff
@@ -36,6 +36,9 @@ class InternalAPIError():
 class SyncError(t: Throwable? = null):
     BaseException(Exception(t), "Session sync failed, check your network connection.")
 
+class SessionExportFailedError(t: Throwable? = null):
+        BaseException(Exception(t), "You need to have internet connection to export session data")
```
lesniakania, line 40: You are using this in both connection error and just an API error, so maybe you should have some more generic message here. Maybe something like "Something wrong happened during exporting session data.". We can change it later if needed.

PR #112, opened 2021-01-12
Microphone settings branch

michalrazny2:
https://trello.com/c/t55TsPGM/1079-settings-microphone-maximum


lesniakania (approved):

app/src/main/java/io/lunarlogic/aircasting/lib/Settings.kt
```diff
@@ -82,6 +92,10 @@ open class Settings(mApplication: Application) {
         return sharedPreferences.getBoolean(key, default)
     }
 
+    open fun getIntFromSettings(key: String, default: Int): Int { //todo: without default value? why sharedPreferences.getInt() requires normal Int, cant take Int?
```
lesniakania, line 95: Yeah, it is kind of weird why it differs from the `getString` implementation. When you see the source code - there is `@Nullable` annotation in `getString` but not in `getInt`.
I guess you can remove this comment, right?

PR #113, opened 2021-01-13
Time stamps trial

michalrazny2:
https://trello.com/c/ygTO0ZC7/1089-v149-fixed-following-timestamps-are-inaccurate

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
@@ -257,13 +257,16 @@ class Session(
 
         if (endTime == null) return durationString
 
-        if (endTime!!.date != startTime.date && endTime!!.month != startTime.month && endTime!!.year != startTime.year) {
+        if (checkIfTheSameDay(startTime, endTime!!)) {
             durationString += " - ${dateFormatter.format(endTime)} ${hourFormatter.format(endTime)}"
         } else {
             durationString += "-${hourFormatter.format(endTime)}"
         }
-
         return durationString
+    }
+
+    fun checkIfTheSameDay(startTime: Date, endTime: Date): Boolean {
+        return endTime.date != startTime.date && endTime.month != startTime.month && endTime.year != startTime.year
     }
 
     fun infoString(): String {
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
```diff
@@ -257,7 +257,7 @@ class Session(
 
         if (endTime == null) return durationString
 
-        if (endTime!!.day != startTime.day) {
+        if (endTime!!.date != startTime.date && endTime!!.month != startTime.month && endTime!!.year != startTime.year) {
```
lesniakania, line 260: Not sure, but do you mean to change day to date here? Also, it might be good to extract this to a separate method now.

michalrazny2: It seems strange, but indeed- the change from day to date is intentional here. If my research is correct "day" returns us which day of the week is the day, and "date" returns which day of the month is the day.

app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
```diff
@@ -266,6 +266,10 @@ class Session(
         return durationString
     }
 
+    fun checkIfTheSameDay(startTime: Date, endTime: Date): Boolean {
```
mturnau, line 269: i suggest to name it isSameDay and return true if it is the same day

mturnau: also, maybe
```
 val dateFormat = SimpleDateFormat("yyyyMMdd")
dateFormat.format(startTime) == dateFormat.format(endTime)
```
is slightly more readible?

app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
```diff
@@ -257,7 +257,7 @@ class Session(
 
         if (endTime == null) return durationString
 
-        if (endTime!!.day != startTime.day) {
+        if (checkIfTheSameDay(startTime, endTime!!)) {
```
mturnau, line 260: `!isSameDay(startTime, endTime!!)`
or even better, do not use negation and swap the conditions

PR #115, opened 2021-01-15
Refresh before edit

michalrazny2:
https://trello.com/c/f3CDFYlc/1069-edit-session-refresh-the-session-before-showing-edit-screen

I had two major doubts:
- The loader is probably drawn incorrect screen
- I am not sure how to check if this solution really works...


app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionDownloadService.kt
```diff
@@ -27,6 +27,8 @@ class SessionDownloadService(private val apiService: ApiService, private val err
                 } else {
                     errorHandler.handle(UnexpectedAPIError())
                 }
+                 finallyCallback?.invoke() //todo: is this right place for this invoke
```
lesniakania, line 30: Yes, here and also `onFailure` too. Good catch, we forgot to use this ; p

lesniakania (approved):

PR #118, opened 2021-01-19
Forgot password

michalrazny2:
Basic functionality is working I guess. 

[ ] handling errors,
[ ] UI change to show "Profile name or email"
[ ] clean up?


app/src/main/java/io/lunarlogic/aircasting/screens/login/LoginViewMvcImpl.kt
```diff
@@ -21,6 +21,11 @@ class LoginViewMvcImpl : BaseObservableViewMvc<LoginViewMvc.Listener>, LoginView
         createAccountButton?.setOnClickListener {
             onCreateAccountClicked()
         }
+
+        val forgotPasswordButton = rootView?.findViewById<Button>(R.id.forgot_password_button)
```
lesniakania, line 25: we should try to use this magic kotlin thing like `import kotlinx.android.synthetic.main.more_info_bottom_sheet.view.*` in new places

lesniakania (approved):

PR #119, opened 2021-01-19
Stopping download/sync services on logout.

mturnau:
OK, by mistake this PR is fixing two things:
1. LOGOUT - The idea was to stop SyncService and FixedMeasurementsDownloadService before logging out and prevent doing DB inserts when the calls were canceled
2. Background/foreground moving - it should fix the behavior of syncing (app in background - should not sync, eg. when we regain connectivity)


app/src/main/java/io/lunarlogic/aircasting/screens/settings/myaccount/MyAccountController.kt
```diff
@@ -25,12 +25,18 @@ class MyAccountController(
     }
 
     override fun onSignOutClicked() {
+        val event = LogoutEvent()
+        EventBus.getDefault().post(event)
+
         mSettings.logout()
+        SessionsSyncService.cancel()
```
lesniakania, line 32: I'd consider combining `cancel` and `destroy` into one method

app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -92,9 +98,14 @@ class SessionsSyncService {
...
+                    val error = if (call.isCanceled()) {
+                        SyncCanceledError(t)
+                    } else {
+                        SyncError(t)
+                    }
                     syncStarted.set(false)
                     hideLoaderCallback?.invoke()
-                    errorHandler.handleAndDisplay(SyncError(t))
+                    errorHandler.handleAndDisplay(error)
                 }
```
lesniakania, line 109: Not sure if we want log to Crashlytics and display an error when sync was cancelled. I think it should just fail gracefully.

app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -66,9 +72,9 @@ class SessionsSyncService {
             val sessions = sessionRepository.finishedSessions()
             val syncParams = sessions.map { session -> SyncSessionParams(session) }
             val jsonData = gson.toJson(syncParams)
-            val call = apiService.sync(SyncSessionBody(jsonData))
+            call = apiService.sync(SyncSessionBody(jsonData))
 
-            call.enqueue(object : Callback<SyncResponse> {
+            call!!.enqueue(object : Callback<SyncResponse> {
```
lesniakania, line 77: I know, I'm boring with that, but maybe `?` instead of `!!` ;]?

mturnau: Well.. call is nullable because it is initialized by null. But two lines after we assign non nullable Call object maybe !! is actually better because it indicates it will happen for sure? Not sure what is preferable convention though.

app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -50,12 +57,17 @@ class SessionsSyncService {
...
             mSingleton = null
         }
     }
 
     fun sync(showLoaderCallback: (() -> Unit)? = null, hideLoaderCallback: (() -> Unit)? = null) {
-        if (syncStarted.get()) {
+        println("MARYSIA: wants to sync but checking if in background ${syncInBackground.get()}")
+        if (syncInBackground.get()) {
+            triedToSyncBackground.set(true)
+        }
```
mturnau, line 69: this will happen if we regain connectivity when app is in background. When in foreground again, it should sync

app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -124,15 +136,43 @@ class SessionsSyncService {
...
+            errorHandler.handleAndDisplay(SyncError(t))
+        }
+    }
+
+    fun onAppToForeground() {
+        syncInBackground.set(false)
+        if(triedToSyncBackground.get()) {
+            triedToSyncBackground.set(false)
+            sync()
+        }
```
mturnau, line 171: This is the continuation of the logic described in revious comment. I don't like the code very much, any suggestions?

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
```diff
@@ -65,6 +66,11 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
         deleteSession(event.sessionUUID)
     }
 
+    @Subscribe
+    fun onMessageEvent(event: LogoutEvent) {
+        fixedSessionDownloadMeasurementsService.stop()
+    }
```
mturnau, line 72: hm now I'm thinking we use event for this but not for the sync service (becuase it is a singleton we can just stp it from there). Works ok, not sure if we should make it consistent

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
```diff
@@ -86,11 +92,15 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
     }
 
     fun onAppToForeground() {
+        println("MARYSIA: Fixed session download resume")
         fixedSessionDownloadMeasurementsService.resume()
+        sessionsSyncService.onAppToForeground()
```
mturnau, line 97: also, small inconsistency. We make SessionManager responsible for knowing what to do with the fixed service but for syncservice we leave the knowledge to it. Hm.

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
lesniakania: good idea

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
lesniakania: I love your comments <3 : D

PR #122, opened 2021-01-22
disabled maps flow

michalrazny2:
https://trello.com/c/2N9QL7HX/1078-disable-maps-mobile-session-flow-update-screens-text

Doubts:
- possibly a bit of a mess in architecture,
- possibly this is not effective way to do this
- this linear layout surrounding map fragment probably not needed in the end


app/src/main/java/io/lunarlogic/aircasting/screens/new_session/confirmation/ConfirmationViewMvcImpl.kt
```diff
@@ -65,7 +81,7 @@ abstract class ConfirmationViewMvcImpl: BaseObservableViewMvc<ConfirmationViewMv
         googleMap ?: return
         mMap = googleMap
 
-        val sessionLocation = session!!.location!!
+        val sessionLocation = session!!.location!! // todo : tutaj uwaga?
```
lesniakania, line 84: you can change it to `val sessionLocation = session?.location ?: return`

lesniakania (approved):

PR #123, opened 2021-01-22
Fixed sessions should be send to the backend during sync too

lesniakania:


lesniakania:
Closing this without merging after we agreed that we don't allow to create fixed sessions without internet connection

PR #124, opened 2021-01-25
Dashboard default empty screens changes

michalrazny2:
https://trello.com/c/H67E7aoQ/1071-dashboard-default-screen-should-match-proto

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsViewMvcImpl.kt
@@ -31,7 +31,7 @@ abstract class SessionsViewMvcImpl<ListenerType>: BaseObservableViewMvc<Sessions
         this.rootView = inflater.inflate(R.layout.fragment_sessions_tab, parent, false)
 
         mEmptyView = rootView?.findViewById(layoutId())
-        mRecordSessionButton = rootView?.findViewById(R.id.dashboard_record_new_session_button)
+        mRecordSessionButton = rootView?.findViewById(recordNewSessionButtonId())
         mRecordSessionButton?.setOnClickListener {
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/fixed/FixedViewMvcImpl.kt
```diff
@@ -43,4 +44,12 @@ class FixedViewMvcImpl(
...
         }
     }
+
+    override fun layoutId(): Int {
+        return R.id.empty_dashboard
+    }
+
+    override fun recordNewSessionButtonId(): Int {
+        return R.id.dashboard_record_new_session_button
+    }
```
lesniakania, line 54: I'm not 100% sure, but I think you could just give these buttons the same ids - only one is visible at a time. You can check it out. It would simplify a bit - you'd not need this method then.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsViewMvcImpl.kt
```diff
@@ -30,8 +30,8 @@ abstract class SessionsViewMvcImpl<ListenerType>: BaseObservableViewMvc<Sessions
     ): super() {
         this.rootView = inflater.inflate(R.layout.fragment_sessions_tab, parent, false)
 
-        mEmptyView = rootView?.findViewById(R.id.empty_dashboard)
-        mRecordSessionButton = rootView?.findViewById(R.id.dashboard_record_new_session_button)
+        mEmptyView = rootView?.findViewById(layoutId())
+        mRecordSessionButton = rootView?.findViewById(recordNewSessionButtonId())
```
lesniakania, line 34: I think it would be good to change`findViewById` to this magic syntax when we touch them

lesniakania (approved):

PR #134, opened 2021-01-28
Enhance validation on the new/edit/share session views

michalrazny2:
https://trello.com/c/B5KCghZr/1099-enhance-validation-on-the-new-edit-share-session-views


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ShareSessionBottomSheet.kt
```diff
@@ -80,10 +84,18 @@ class ShareSessionBottomSheet(
...
+        if (!ValidationHelper.isValidEmail(emailInput)){
+            showError()
+            return
+        }
         mListener.onShareFilePressed(session, emailInput)
         dismiss()
     }
 
+    private fun showError() {
+        emailInputLayout?.error = " "
```
mturnau, line 96: I'm thinking.. in this case it may not be obvious what is it about. Especially, if someone for any reason has weird email that won't go through our validation, they won't know what is it about. Try showing a message, however, if it looks very bad let's leave it as it is.

PR #136, opened 2021-01-28
Streams delete

yzmopolis:


app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
```diff
@@ -157,6 +157,7 @@ class Session(
     val indoor get() = mIndoor
     val streamingMethod get() = mStreamingMethod
     val followed get() = followedAt != null
+    val activeStreams get() = mStreams.filter { stream -> !stream.deleted }
```
yzmopolis, line 160: I've added this in case there is no internet connection, but I'm not sure if it's necessary 🤔

app/src/androidTest/java/io/lunarlogic/aircasting/MobileSessionTest.kt
```diff
@@ -240,9 +241,15 @@ class MobileSessionTest {
...
+            .check(matches(isNotChecked()))
+            .perform(click())
+            .check(matches(isChecked()))
+
+        onView(withId(R.id.delete_streams_button)).perform(click())
         Thread.sleep(2000)
         // check if session deleted
-        onView(withText("Ania's mobile mic session")).check(matches(not(isDisplayed())))
-
+        onView(withText("Ania's mobile mic session")).check(doesNotExist())
```
yzmopolis, line 253: It asserts that a view is not present.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
```diff
@@ -0,0 +1,116 @@
...
+        fun onDeleteStreamsPressed(session: Session)
+    }
+    private var mStreamsOptionsContainer: LinearLayout? = null
+    private var checkBoxMap: HashMap<CheckBox, DeleteStreamOption> = HashMap()
+
+    class DeleteStreamOption(
+        val allStreamsBoxSelected: Boolean,
+        val stream: MeasurementStream? = null
+    )
+
```
yzmopolis, line 32: Handling checkboxes was easier on hashmaps. 
I've created a class for checkbox options:
`class DeleteStreamOption(
        val allStreamsBoxSelected: Boolean,
        val stream: MeasurementStream? = null
    )`
For checkbox `All from this session` val `allStreamsBoxSelected` is truthy and there is no streams. For checkboxes with streams val `allStreamsBoxSelected` is falsy and it have streams. I didn't have better idea for this, if you have any suggestions just let me know.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
```diff
@@ -0,0 +1,116 @@
...
+        return view
+    }
+
+    fun getSelectedValues(): List<MeasurementStream> {
+        val streamsSelectedForDeletion: MutableList<MeasurementStream> = ArrayList()
+        selectedStreams().forEach { option ->
+            option.stream?.let { streamsSelectedForDeletion.add(it) }
+        }
+        return Collections.unmodifiableList(streamsSelectedForDeletion)
+    }
```
yzmopolis: Returns list of streams get from the selected options.

yzmopolis (commented):
I've added my comments and questions :)

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
```diff
@@ -0,0 +1,116 @@
...
+        return view
+    }
+
+    fun getStreamsToDelete(): List<MeasurementStream> {
+        val streamsSelectedForDeletion: MutableList<MeasurementStream> = ArrayList()
+        selectedOptions().forEach { option ->
+            option.stream?.let { streamsSelectedForDeletion.add(it) }
+        }
+        return Collections.unmodifiableList(streamsSelectedForDeletion)
+    }
```
yzmopolis, line 70: Returns a list of streams to delete from selected options.

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
```diff
@@ -200,4 +206,27 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
...
+
+    private fun deleteStreams(session: Session, streamsToDelete: List<MeasurementStream>?) {
+        markForRemoval(session, streamsToDelete) {
+            val reloadedSession = sessionsRespository.loadSessionAndMeasurementsByUUID(session.uuid)
+            if (reloadedSession != null) {
+                sessionUpdateService.update(reloadedSession) {
+                    DatabaseProvider.runQuery {
+                        measurementStreamsRepository.deleteMarkedForRemoval()
+                        sessionsSyncService.sync()
+                    }
```
yzmopolis, line 218: I was thinking about divide it into smaller methods, what do you think?

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
lesniakania: You could use `mapNotNull` or a combination of `map` and `filterNotNull()` afterward, to simplify this method

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/fixed/FixedController.kt
```diff
@@ -49,16 +50,38 @@ class FixedController(
...
+        deleteSessionDialog?.dismiss()
+    }
+
+    private fun deleteStreams(session: Session, streamsToDelete: List<MeasurementStream>?) {
+        val event = DeleteStreamsEvent(session, streamsToDelete)
+        EventBus.getDefault().post(event)
+        deleteSessionDialog?.dismiss()
+    }
+
+    private fun deleteAllStreamsSelected(allStreamsBoxSelected: Boolean, selectedOptionsCount: Int?, sessionStreamsCount: Int?): Boolean {
```
lesniakania, line 83: It's not a strong opinion, but maybe it would be more elegant to have two methods in the listener - one for deleting all streams and one for deleting particular streams. So basically handle what was clicked in the view and then in the controller just react to that.

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
lesniakania: yeah, it's a good idea

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
lesniakania: I think I'd put the one for deleting all streams, outside this map, but it's not something super important, I guess.
Having this class + a map of checkboxes for the rest is a good choice though.

app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
lesniakania: Yes, I think it makes sense to have it. You can create a mobile session without an internet connection, so you can also delete sessions from it.

lesniakania (approved):

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
yzmopolis: @lesniakania I think I don't understand what do you mean exactly, how it could simplify this method?

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
yzmopolis: Yeah, I was thinking about this, but it was less intuitive for me because the checkbox with all streams selected would look like this:
`checkBoxMap[checkbox] = DeleteStreamOption()` and it should have its own observer, but maybe there is something I missed?

lesniakania: Do you mean listener, yeah, although I'm not sure if you are using this listener anyway, so it will be even simpler.
This will simplify `allStreamsBoxSelected` - you can just check if this particular separate checkbox is checked.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
lesniakania: Sorry, you need `map`

```kotlin
fun getStreamsToDelete(): List<MeasurementStream> {
    return selectedOptions().mapNotNull { option -> option.stream }
}
```

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
```diff
@@ -0,0 +1,108 @@
...
+            mStreamsOptionsContainer?.addView(streamCheckboxView)
+        }
+    }
+
+    private fun createCheckboxView(checkbox: CheckBox, displayedValue: String?): View {
+        checkbox.id = View.generateViewId()
+        checkbox.text = displayedValue
+        val layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT))
+        checkbox.layoutParams = layoutParams
+        checkbox.buttonTintList = ColorStateList.valueOf(resources.getColor(R.color.aircasting_blue_400))
```
mturnau, line 103: As I already mentioned on slack, you can, instead of putting tint color here, set drawable `checkbox.buttonDrawable = drawable` and place correct vectors for checked/unchecked state. But you can do it in separate PR as it is a separate task now

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -161,6 +162,15 @@ abstract class SessionsController(
         startShareSessionBottomSheet(session)
     }
 
+    override fun onDeleteSessionClicked(session: Session){
```
mturnau, line 165: missing space before opening mustache ;)

app/src/main/res/values/styles.xml
```diff
@@ -210,6 +210,28 @@
...
+    </style>
+
+    <style name="TextAppearance.Aircasting.Checkbox" parent="@style/TextAppearance.Aircasting.Body1">
+        <item name="android:textAlignment">center</item>
+        <item name="lineHeight">32dp</item>
+        <item name="android:fontFamily">@font/muli_medium</item>
+        <item name="android:textColor">@color/aircasting_grey_300</item>
+        <item name="android:button">@drawable/checkbox_selector</item>
+        <item name="android:textSize">@dimen/text_size_m</item>
+        <item name="materialThemeOverlay">@style/ThemeOverlay.Aircasting.Checkbox</item>
```
mturnau, line 228: I'm not an expert in theme overlays but I'm not sure why do we need this here. Was it the ony way to make the text gray?

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
yzmopolis: https://www.youtube.com/watch?v=PvUcNMZJx5k 😆

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
yzmopolis: I'll do it in a separate task  :)

app/src/main/res/values/styles.xml
yzmopolis: you're right, we don't need it here

PR #145, opened 2021-02-03
Fixed checking location permissions on sync from SD card triggered

lesniakania:
On older Android OSes, checking permissions with checkSelfPermission does not work properly


app/src/main/AndroidManifest.xml
```diff
@@ -8,7 +8,6 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
-    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
```
lesniakania, line 11: I realized it was already there, in 6th line

PR #146, opened 2021-02-03
Smaller map graph follow unfollow buttons

yzmopolis:
Important changes:

- I've changed paddings and dimens for `follow`, `unfollow`, `map`, `graphs` buttons
- I've added containers for buttons
- I used TouchDelegate to keep the touchable area the same as earlier 

Screenshot 2021-02-03 at 19 18 39

AC:

- [x] it matches proto: https://projects.invisionapp.com/share/DCWT9W6BUE7#/screens/414476845
- [ ] touchable area is the same

[TouchDelegate docs](https://developer.android.com/reference/android/view/TouchDelegate.html)

[ticket](https://trello.com/c/3fb7fnUc/1070-cards-map-graph-follow-buttons-should-match-the-proto)

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
@@ -1,10 +1,13 @@
 package io.lunarlogic.aircasting.screens.dashboard
 
+import android.graphics.Rect
 import android.view.LayoutInflater
+import android.view.TouchDelegate
 import android.view.View
 import android.view.ViewGroup
 import android.widget.Button
 import android.widget.ImageView
+import android.widget.LinearLayout
 import android.widget.TextView
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.fragment.app.FragmentManager
@@ -43,6 +46,11 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
     private var mMapButton: Button
     private var mGraphButton: Button
     private var mLoader: ImageView?
+
+    private var mFollowTouchableArea: LinearLayout
+    private var mUnfollowTouchableArea: LinearLayout
+    private var mMapTouchableArea: LinearLayout
+    private var mGraphTouchableArea: LinearLayout
 
     protected var mSessionPresenter: SessionPresenter? = null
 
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/dimens.xml
```diff
@@ -46,6 +46,10 @@
 
     <dimen name="card_corner_radius">0dp</dimen>
 
+    <dimen name="height_0">0dp</dimen>
+    <dimen name="padding_2">2dp</dimen>
+    <dimen name="margin_top_5">5dp</dimen>
```
lesniakania, line 51: when we started implementing layouts with @mturnau we used similar constants, but at some point, we agreed (I think after reading some article, but I don't remember exactly) that when we have a situation that we just need to use specific dimensions, we are just using it in the XML as it is, without adding it to the dimens file.
I guess it was about readability and not making too many and a bit artificial dimens.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -91,25 +99,34 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
             onCollapseSessionCardClicked()
             collapseSessionCard()
         }
+
         mFollowButton = findViewById(R.id.follow_button)
+        mFollowTouchableArea = findViewById(R.id.follow_container)
```
lesniakania, line 104: I wanted to comment that we agreed on using "the new magic syntax" as I was referring to kotlin synthetics syntax, but I wanted to google it and look what I just found....:
https://medium.com/better-programming/why-are-kotlin-synthetics-deprecated-and-what-are-the-alternatives-5c2b087dda1c /cc @michalrazny2 ....
I guess we should be moving to the data binding syntax, but let's not worry about this now, maybe just when you create any new views, please use that.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -44,6 +47,11 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
     private var mGraphButton: Button
     private var mLoader: ImageView?
 
+    private var mFollowTouchableArea: LinearLayout
+    private var mUnfollowTouchableArea: LinearLayout
+    private var mMapTouchableArea: LinearLayout
+    private var mGraphTouchableArea: LinearLayout
```
lesniakania, line 53: I'm a bit concerned with these linear layouts - too many layouts (can cause performance problems)[https://stackoverflow.com/a/15405309] and we have lots of sessions on the fixed and dormant tabs to display.
I'd try removing them and in the `expandViewHitArea` use something like this:
```kotlin
childRect.top -= TOUCHABLE_PADDING;
childRect.left -= TOUCHABLE_PADDING;
childRect.right += TOUCHABLE_PADDING;
childRect.bottom += TOUCHABLE_PADDING;
```

where `TOUCHABLE_PADDING` would be some constant defined in this class.
you can check more examples [here](https://www.programcreek.com/java-api-examples/?api=android.view.TouchDelegate).

lesniakania (changes requested):

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
yzmopolis: Thanks for this comment, I'll remember it :)

app/src/main/res/values/dimens.xml
yzmopolis: @lesniakania of course, I didn't know, I'll follow your approach 🐾

lesniakania (approved):

PR #149, opened 2021-02-05
Did you know box

michalrazny2:
https://trello.com/c/FUDTWZVX/1100-mobile-dormant-mobile-active-no-sessions-to-list-add-did-you-know-box


app/src/main/res/values/colors.xml
```diff
@@ -23,4 +23,5 @@
     <color name="mtrl_textinput_default_box_stroke_color">@color/aircasting_grey_100</color>
 
     <color name="graph_overlay">#40FFFFFF</color>
+    <color name="did_you_know_box_color">#F2F8FF</color>
```
lesniakania, line 26: I think we should not introduce yet another shade of gray ;) But let's leave it for now, I added it to my list of things to talk through with Pina :]

PR #164, opened 2021-02-15
Display name for a bluetooth device should be the same as in Settings -> Bluetooth

lesniakania:
[Task](https://trello.com/c/1aTwHQFB/1037-change-the-app-to-use-the-esp32-device-name-rather-than-the-ble-mac-address)


yzmopolis (approved):
Looks good! :shipit:

PR #166, opened 2021-02-15
Session details - gain more vertical space for map & graph

mturnau:
The [ticket](https://trello.com/c/FoUxrECt/971-map-graph-adjust-spacing-to-gain-vertical-display-space) was created because sessiin details looked very bad on smallest screens (example was Motorola with 720x1280 resolution). Micheal added few ideas, after consulting with Pina we decided to make a few changes (please review subtasks in the [ticket](https://trello.com/c/FoUxrECt/971-map-graph-adjust-spacing-to-gain-vertical-display-space) )

Additionally, I allowed myself to introduce qualifiers. For bigger screens (height > 600dp) I left the measurements description and the "more" button. For smaller screens they are hidden. I made hlu slider labels area clickable (it's like more button) so on smaller screens you can still get edit HLU thresholds. Both the min height (now 600 dp) and the elements can be changed, these are just my propositions. To be reviewed with Michael and/or Pina. 

Also, the statistics box (avg/now/peak) always takes the same percent of the with, so  on smaller screens it will be shrinked (smaller spaces between the values)

See screenshots for small vs big screen.

Small (< 600dp)
Screenshot 2021-02-15 at 12 17 05

Screenshot 2021-02-15 at 12 16 40

Medium (> 600dp)

Screenshot 2021-02-15 at 12 32 06
Screenshot 2021-02-15 at 12 32 31

Big (> 600dp)
Screenshot_20210215-123411
Screenshot_20210215-123429

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
@@ -4,6 +4,7 @@ import android.location.Location
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.Button
 import android.widget.ImageView
 import android.widget.TextView
 import androidx.fragment.app.FragmentManager
@@ -34,6 +35,7 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
     private val mMeasurementsTableContainer: MeasurementsTableContainer
     protected var mStatisticsContainer: StatisticsContainer?
     private val mMoreButton: ImageView?
+    private val mMoreInvisibleButton: Button?
     private val mHLUSlider: HLUSlider
 
     constructor(
@@ -47,7 +49,7 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
 
         mSessionDateTextView = this.rootView?.session_date
         mSessionNameTextView = this.rootView?.session_name
-        mSessionMeasurementsDescription = this.rootView?.session_measurements_description
+        mSessionMeasurementsDescription = this.findViewById(R.id.session_measurements_description)
 
         mMeasurementsTableContainer = MeasurementsTableContainer(
             context,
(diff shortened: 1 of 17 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
```diff
@@ -93,7 +97,11 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
     }
 
     private fun showSlider() {
-        mMoreButton?.visibility = View.VISIBLE
+        if (context.resources.getInteger(R.integer.visible_in_larger_screens) != 0) {
```
mturnau, line 100: I know it is weird. The problem is, the integer I have in integers.xml is 0,1,2 (whoch is index of the actual View enum) because this is how it is used in XMLs. Here I can's use it directly though, programmatically you need to assign the actual values (0,4,8). Any suggestions how to make this better welcome.

lesniakania: I'm not sure if I fully understand how it works, but maybe you can try adding `boolean` instead of `integer` for `visible_in_larger_screens` there? And set it to `false` on the smaller screens xml.
Also, do you use `view_visible`/`view_invisible`/`view_gone` from this file?

lesniakania (approved):

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
mturnau: @lesniakania Yeah but then I can't directly use booleans in XML (it would work only in the code)
view_ are used to set `visible_in_larger_screens` in a meaningful way,

lesniakania: Ah, ok, now I get it. So how about this?

```kotlin
val visibility = context.resources.getInteger(R.integer.visible_in_larger_screens)
if (visibility != View.VISIBLE) {
  //...
}
```

oooor, hm - would this work? If yes, I think this would be better, without the `if` ;)

```kotlin
val visibility = context.resources.getInteger(R.integer.visible_in_larger_screens)
mMoreButton?.visibility = visibility
```

lesniakania: Eh, ok, I see it won't work, because this values in xmls are indexes not actual visibility. Omg, Android, eh.
So nevermind, it's not that big deal. Maybe I'd name it with `index` suffix though and add a comment, for more clarity.

```kotlin
val visibilityIndex = context.resources.getInteger(R.integer.visible_in_larger_screens)
// 0 means visible, see VISIBILITY_FLAGS in View class:
// private static final int[] VISIBILITY_FLAGS = {VISIBLE, INVISIBLE, GONE};
if (visibilityIndex != 0) {
  //...
}
```

PR #169, opened 2021-02-16
Onboarding

michalrazny2:
https://trello.com/c/Dq83XRSJ/831-onboarding


app/src/main/res/layout/onboarding_page_1.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <ImageView
```
mturnau, line 7: This image is supposed to take full width. I'm afraid the easiest way would be to remove paddings from onboarding_activity and add them to all elements individually (then you can have 0 paddings on that image). Playing with negative paddings is rather the last resort. Maybe there is some other way but couldn't find it easily.

app/src/main/java/io/lunarlogic/aircasting/screens/onboarding/OnboardingWizardNavigator.kt
```diff
@@ -0,0 +1,93 @@
...
+    fun goToPage2(listener: OnboardingPage2ViewMvc.Listener) {
+        incrementStepProgress()
+        val fragment = OnboardingPage2Fragment()
+        fragment.listener = listener
+        goToFragment(fragment)
+    }
+
+    fun goToPage3(listener: OnboardingPage3ViewMvc.Listener) {
+        incrementStepProgress()
+        val fragment = OnboardingPage3Fragment()
```
mturnau, line 46: not sure in which place but when everything is green, the progress bar color should be green as well. There are several ways to do it. A hint would be (maybe) that primaryColor is used there in XML so maybe you could apply style overlay only in that view to change primary color?

app/src/main/res/layout/activity_create_account.xml
```diff
@@ -3,11 +3,41 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:fillViewport="true">
+    android:fillViewport="true"
+    android:background="@drawable/dashboard_background">
```
mturnau, line 7: is this a monochromatic rainbow thingy? If so, note that in create account it should be blue, not green!

mturnau: ah and also it is green on the dashboard.. should be blue also there

app/src/main/res/layout/onboarding_page_2.xml
```diff
@@ -0,0 +1,75 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:id="@+id/guidelineTop"
+        app:layout_constraintGuide_percent="0.1"
+        android:orientation="horizontal"/>
+
+    <ImageView
+        android:id="@+id/onboarding_step_2_image"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
```
mturnau, line 17: It is super important that when you want to constraint this image to guidelines, you need to set height to 0dp (match constraints) and width to wrap_content. Otherwise it won't work as intended. I suggest you change it in all screens with picture and text and check on different sizes of screen. PRO TIP: Using an emulator is not needed in fact. Go to the xml file, select design view and you can select the device on the top left. If you scroll to the bottom of the menu you have "generic phones and tablets" and you have all sizes, like 3,2'' as well.

app/src/main/java/io/lunarlogic/aircasting/screens/create_account/CreateAccountViewMvcImpl.kt
```diff
@@ -23,6 +26,13 @@ class CreateAccountViewMvcImpl : BaseObservableViewMvc<CreateAccountViewMvc.List
         loginButton?.setOnClickListener {
             onLoginClicked()
         }
+
+        if (settings.shouldOnboardingAppear()) {
+            val progressBar = rootView?.findViewById<ProgressBar>(R.id.progress_bar)
+            progressBar?.visibility = View.VISIBLE
+            progressBar?.progress = 75   // todo: hardcoded for now <?>
```
lesniakania, line 33: yeah I guess hardcoded is fine, but maybe just in the XML directly?

lesniakania: and sign in view also has progress bar

app/src/main/java/io/lunarlogic/aircasting/screens/onboarding/OnboardingController.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.lunarlogic.aircasting.screens.onboarding.page3.OnboardingPage3ViewMvc
+import io.lunarlogic.aircasting.screens.onboarding.page4.LearnMorePage4BottomSheet
+import io.lunarlogic.aircasting.screens.onboarding.page4.OnboardingPage4ViewMvc
+
+class OnboardingController(
+    private val mContextActivity: AppCompatActivity,
+    mViewMvc: OnboardingViewMvc,
+    private val mFragmentManager: FragmentManager,
+    private val mSettings: Settings
+): OnboardingPage1ViewMvc.Listener,
```
lesniakania, line 19: I'd name those steps more verbosly - e.g `OnboardingGetStarted`, `OnboardingHowsTheAir` etc

app/src/main/java/io/lunarlogic/aircasting/screens/onboarding/OnboardingWizardNavigator.kt
```diff
@@ -0,0 +1,93 @@
...
+import io.lunarlogic.aircasting.screens.onboarding.page1.OnboardingPage1Fragment
+import io.lunarlogic.aircasting.screens.onboarding.page1.OnboardingPage1ViewMvc
+import io.lunarlogic.aircasting.screens.onboarding.page2.OnboardingPage2Fragment
+import io.lunarlogic.aircasting.screens.onboarding.page2.OnboardingPage2ViewMvc
+import io.lunarlogic.aircasting.screens.onboarding.page3.OnboardingPage3Fragment
+import io.lunarlogic.aircasting.screens.onboarding.page3.OnboardingPage3ViewMvc
+import io.lunarlogic.aircasting.screens.onboarding.page4.OnboardingPage4Fragment
+import io.lunarlogic.aircasting.screens.onboarding.page4.OnboardingPage4ViewMvc
+
+class OnboardingWizardNavigator(
```
lesniakania, line 18: as agreed on the call, add base wizard on this branch

lesniakania (changes requested):

lesniakania:
fixe tests by setting `shouldOnboardingAppear` setting and add additional test for onboarding flow

app/src/main/res/layout/activity_create_account.xml
lesniakania: and should be more transparent

app/src/main/res/layout/onboarding_page_1.xml
michalrazny2: isn't it enough if I just set width as "match_parent" ?

michalrazny2: the problem is I have no paddings in activity_onboarding now :|

lesniakania (approved):

PR #172, opened 2021-02-17
Hidden map button when indoor session

yzmopolis:
I've added the `bindMapButton` function in order to hide the `map button when it's an indoor session - both for Fixed and Following sessions.

Screenshot 2021-02-17 at 18 40 54

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/fixed/FixedSessionViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/fixed/FixedSessionViewMvcImpl.kt
@@ -52,6 +52,12 @@ class FixedSessionViewMvcImpl(
         }
     }
 
+    override fun bindMapButton(sessionPresenter: SessionPresenter) {
+        if (sessionPresenter.session?.indoor == true ) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/fixed/FixedSessionViewMvcImpl.kt
```diff
@@ -52,6 +52,12 @@ class FixedSessionViewMvcImpl(
         }
     }
 
+    override fun bindMapButton(sessionPresenter: SessionPresenter) {
+        if (sessionPresenter.session?.indoor == true ) {
+            mMapButton.visibility = View.GONE
+        }
+    }
```
lesniakania, line 59: I think you can move this code to the `SessionViewMvcImpl`, to not have it duplicated in two places. I think you don't need to override anything in the subclasses then because mobile sessions should not have indoor either way, but please double-check ;]

yzmopolis: @lesniakania, yeah, you're right, good catch :)

PR #177, opened 2021-02-23
Delete streams or session confirmation dialog

yzmopolis:
**Important changes**:
-I've added a custom alert dialog layout
-I've added an alert dialog before deleting streams or sessions

AC:
- [x] On click `delete streams` confirmation dialog should appear
- [x] On click `ok` button should delete streams or session
- [x] On click `cancel` should come back to delete session dialog

Screenshot 2021-02-23 at 16 28 14
 [ticket](https://trello.com/c/M0xQzlU9/1138-add-confirmation-dialog-before-deleting-a-session)

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ConfirmationDeleteSessionDialog.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ConfirmationDeleteSessionDialog.kt
@@ -1,0 +1,30 @@
+package io.lunarlogic.aircasting.screens.dashboard
+
+import android.view.LayoutInflater
+import android.view.View
+import androidx.fragment.app.FragmentManager
+import io.lunarlogic.aircasting.R
+import io.lunarlogic.aircasting.screens.common.BaseDialog
+import kotlinx.android.synthetic.main.network_password_dialog.view.*
+import kotlinx.android.synthetic.main.session_actions.view.cancel_button
+
+class ConfirmationDeleteSessionDialog(
+    private val mFragmentManager: FragmentManager,
+    val function: () -> (Unit)
+): BaseDialog(mFragmentManager) {
+    private lateinit var mView: View
+    override fun setupView(inflater: LayoutInflater): View {
+        mView = inflater.inflate(R.layout.confirmation_dialog, null)
+        mView.ok_button.setOnClickListener {
+            okButtonClicked()
+        }
+        mView.cancel_button.setOnClickListener {
+            dismiss()
+        }
+        return mView
+    }
+    private fun okButtonClicked() {
+        function()
+        dismiss()
+    }
+}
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -177,9 +181,33 @@ abstract class SessionsController(
...
+        val cancelButton: Button = dialogView.findViewById(R.id.cancel_button) as Button
+        val header: TextView = dialogView.findViewById(R.id.confirmation_dialog_header)
+        header.text = this.context?.getString(R.string.are_you_sure)
+        val dialog: AlertDialog = builder.create()
+        dialog.show()
+
+        okButton.setOnClickListener {
+            function()
+            dialog.dismiss()
+        }
```
yzmopolis, line 206: I couldn't use setButtonPositive etc methods, because there is no option to have these default buttons customized. When we use a custom layout with our buttons it's only possible to add listeners to them. I don't like it, but well... 🦝

lesniakania: I think it might be better to do the same as in `NetworkPasswordDialog` - do a separate class for that and put this logic there, with the listener.

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -177,9 +181,33 @@ abstract class SessionsController(
...
+    private fun showConfirmationDialog(function: () -> (Unit)) {
+        val builder = AlertDialog.Builder(this.context)
+        val inflater = this.mRootActivity?.layoutInflater!!
+        val dialogView: View = inflater.inflate(R.layout.confirmation_dialog, null)
+        builder.setView(dialogView)
+
+        val okButton: Button = dialogView.findViewById(R.id.ok_button) as Button
+        val cancelButton: Button = dialogView.findViewById(R.id.cancel_button) as Button
+        val header: TextView = dialogView.findViewById(R.id.confirmation_dialog_header)
+        header.text = this.context?.getString(R.string.are_you_sure)
```
lesniakania, line 199: I think you can move it to the XML?

lesniakania (changes requested):

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ConfirmationDeleteSessionDialog.kt
```diff
@@ -0,0 +1,30 @@
...
+        mView.ok_button.setOnClickListener {
+            okButtonClicked()
+        }
+        mView.cancel_button.setOnClickListener {
+            dismiss()
+        }
+        return mView
+    }
+    private fun okButtonClicked() {
+        function()
```
lesniakania, line 27: sorry, last neat picking - can you change a name `function` to `okCallback`?

lesniakania (approved):

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/ConfirmationDeleteSessionDialog.kt
yzmopolis: of course! 🐈

PR #188, opened 2021-03-01
Remove old measurements

yzmopolis:
Important changes:

- I've added `removeOldMeasurements()` function during sync session
- I've created `RemoveOldMeasurementsService` in order to find all fixed sessions measurements older than 24 hours 
- I've added getting sessions ids by their type both in data class and repository
- I've added getting measurements by session ids both in data class and repository
- I've added deleting measurements by their id both in data class and repository

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
@@ -170,4 +170,7 @@ interface SessionDao {
 
     @Query("DELETE FROM sessions WHERE uuid in (:uuids)")
     fun delete(uuids: List<String>)
+
+    @Query("SELECT id FROM sessions WHERE type=:type AND deleted=0")
+    fun loadSessionUuidsByType(type: Session.Type): List<Long>
 }
--- a/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
@@ -45,4 +45,9 @@ class MeasurementsRepository {
         val measurement = mDatabase.measurements().lastForStream(sessionId, measurementStreamId)
         return measurement?.time
     }
+
+    fun getMeasurementsBySessionsIds(sessionsIds: List<Long>): List<MeasurementDBObject> {
+        return mDatabase.measurements().getBySessionsIds(sessionsIds)
+    }
+
 }
--- a/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
@@ -46,13 +46,9 @@ class MeasurementsRepository {
         return measurement?.time
     }
 
-    fun getMeasurementsBySessionsIds(sessionsIds: List<Long>): List<MeasurementDBObject> {
-        return mDatabase.measurements().getBySessionsIds(sessionsIds)
-    }
-
-    fun deleteMeasurements(measurementsIds: List<Long>) {
+    fun deleteMeasurements(sessionIds: List<Long>) {
         mDatabase.runInTransaction {
-            mDatabase.measurements().deleteInTransaction(measurementsIds)
+            mDatabase.measurements().deleteInTransaction(sessionIds)
         }
     }
 
--- a/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
@@ -50,4 +50,8 @@ class MeasurementsRepository {
         return mDatabase.measurements().getBySessionsIds(sessionsIds)
     }
 
+    fun deleteMeasurements(measurementsIds: List<Long>) {
+        measurementsIds.forEach { measurementId -> mDatabase.measurements().delete(measurementId) }
+    }
+
 }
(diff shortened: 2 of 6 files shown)


app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
```diff
@@ -45,4 +45,13 @@ class MeasurementsRepository {
         val measurement = mDatabase.measurements().lastForStream(sessionId, measurementStreamId)
         return measurement?.time
     }
+
+    fun getMeasurementsBySessionsIds(sessionsIds: List<Long>): List<MeasurementDBObject> {
+        return mDatabase.measurements().getBySessionsIds(sessionsIds)
+    }
+
+    fun deleteMeasurements(measurementsIds: List<Long>) {
+        measurementsIds.forEach { measurementId -> mDatabase.measurements().delete(measurementId) }
```
yzmopolis, line 54: I wanted to do it in batch (inTransaction) for all ids selected for deletion, but I've got `SQLiteException: too many SQL variables`. I know how to do it in Rails, but I failed when I tried this:
https://developer.android.com/reference/kotlin/androidx/room/RoomDatabase#runintransaction
Do you have any experience with that?

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -0,0 +1,41 @@
...
+            // `measurement.time.time` looks weird, but it's just getTime() current syntax
+            // it returns the number of milliseconds since January 1, 1970, 00:00:00 GMT represented by this Date object.
+            // Source: https://developer.android.com/reference/kotlin/java/util/Date
+            val measurementTimeInMilliseconds = measurement.time.time
+            if (measurementOlderThanTwentyFourHours(measurementTimeInMilliseconds)) {
+                measurementsIdsForDeletion.add(measurement.id)
+            }
+        }
+        return measurementsIdsForDeletion
+    }
```
yzmopolis, line 34: I know I could use `.map` here instead of a mutable list, but it was more readable for me. But if you don't like this solution just let me know :)

yzmopolis:
Btw all tests passed locally 🟢

app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
```diff
@@ -45,4 +45,15 @@ class MeasurementsRepository {
         val measurement = mDatabase.measurements().lastForStream(sessionId, measurementStreamId)
         return measurement?.time
     }
+
+    fun getMeasurementsBySessionsIds(sessionsIds: List<Long>): List<MeasurementDBObject> {
+        return mDatabase.measurements().getBySessionsIds(sessionsIds)
+    }
+
+    fun deleteMeasurements(measurementsIds: List<Long>) {
+        mDatabase.runInTransaction {
```
mturnau, line 54: Really, you need both @transaction and runInTransaction? I'm not saying you don't - it just looks suspicious. However, with Room everything is possible ;)

yzmopolis: To be honest - I'm not sure, but when I tried run it without `@transaction` I've got an error, but maybe there is different approach to fix this, but I didn't find it...

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -0,0 +1,41 @@
...
+        if (measurementsForDeletionIds.isNotEmpty()) {
+            measurementRepository.deleteMeasurements(measurementsForDeletionIds)
+        }
+    }
+
+    private fun getFixedSessionsMeasurements(fixedSessionsIds: List<Long>): List<MeasurementDBObject> {
+        return measurementRepository.getMeasurementsBySessionsIds(fixedSessionsIds)
+    }
+
+    private fun findMeasurementsForDeletion(fixedSessionsMeasurements: List<MeasurementDBObject>): List<Long> {
```
mturnau, line 22: I think you could do it all by SQL as well... sth like:
`DELETE FROM measurements WHERE time <= datetime('now', '-1 day') AND session_id IN ([fixed sessions ids])`

app/src/main/java/io/lunarlogic/aircasting/database/repositories/MeasurementsRepository.kt
mturnau: Try the other way around - with @transaction but without runInTransaction. Just to make sure...

app/src/main/java/io/lunarlogic/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -125,6 +128,11 @@ class SessionsSyncService {
         }
     }
 
+    private fun removeOldMeasurements() {
+        val fixedSessionsIds = sessionRepository.sessionsIdsByType(Session.Type.FIXED)
```
mturnau, line 132: Not sure but maybe we should move also selecting fixed sessions id to the service? To have all of the logic in one place...

app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
```diff
@@ -170,4 +170,7 @@ interface SessionDao {
 
     @Query("DELETE FROM sessions WHERE uuid in (:uuids)")
     fun delete(uuids: List<String>)
+
+    @Query("SELECT id FROM sessions WHERE type=:type AND deleted=0")
+    fun loadSessionUuidsByType(type: Session.Type): List<Long>
```
lesniakania, line 175: I know it's already merged, but please update the name when convenient - uuid is a different field, so I think it should be named `loadSessionIdsByType`

PR #191, opened 2021-03-02
Increase top margin of delete streams button

yzmopolis:
Important changes:

- I've increased the top margin of the delete streams button
- I've changed the font color for checkbox label regardless of the checkbox state
- I've adjusted only font sizes (instead of increasing the bottom sheet) since we have a dedicated ticket for that (and it doesn't help very much)
- I've added a listener in order to check all checkboxes when deleting all is selected

Screenshot 2021-03-03 at 17 17 01

[Ticket](https://trello.com/c/pBhBwiWD/1142-enhancements-for-delete-streams-form)

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
@@ -53,7 +53,24 @@ class DeleteSessionBottomSheet(private val mListener: Listener, private val sess
         mStreamsOptionsContainer = view?.findViewById(R.id.streams_options_container)
         generateStreamsOptions()
 
+        allStreamsCheckboxListener()
         return view
+    }
+
+    private fun allStreamsCheckboxListener() {
+        allStreamsCheckbox.setOnCheckedChangeListener { checkboxView, _ ->
+            if (checkboxView.isChecked) {
+                setAllCheckboxesState(true)
+            } else {
+                setAllCheckboxesState(false)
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
```diff
@@ -53,9 +53,26 @@ class DeleteSessionBottomSheet(private val mListener: Listener, private val sess
...
 
+    private fun allStreamsCheckboxListener() {
+        allStreamsCheckbox.setOnCheckedChangeListener { checkboxView, _ ->
+            if (checkboxView.isChecked) {
+                setAllCheckboxesState(true)
+            } else {
+                setAllCheckboxesState(false)
+            }
+        }
+    }
```
lesniakania, line 68: it's not really blocking comment, but I'd renamed this method to `setAllStreamsCheckboxListener` and do small refactor:

```kotlin
fun setAllStreamsCheckboxListener() {
  allStreamsCheckbox.setOnCheckedChangeListener { checkboxView, _ ->
    setAllCheckboxesState(checkboxView.isChecked)
  }
}
```

lesniakania (approved):

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/DeleteSessionBottomSheet.kt
yzmopolis: @lesniakania done <3

PR #194, opened 2021-03-03
Fix progress bar to show correct progress

michalrazny2:
https://trello.com/c/29pfMpDY/1154-fix-progress-bar-to-show-correct-progress

The idea:
- we got minimum number of steps that we have to go through when creating session (Phone mic session- 3 steps)
- basically we set the maximum value of progress on ProgressBar on 40 (10 per step + 10 because it looks better)
- in NewSessionController we increase the MAX_PROGRESS value for each step we have to add to our flow (+1 if locations is off, +1 if Bluetooth is off, +1 if maps are disabled (turn off location services screen), +3 screens if DeviceType==Bluetto device)

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
@@ -16,7 +16,7 @@ abstract class BaseWizardNavigator(
     }
 
     protected abstract val STEP_PROGRESS: Int
-    val progressBarCounter = ProgressBarCounter(4) // we got 3 basic steps in the flow (progress bar should have 1 more to look better)
+    val progressBarCounter = ProgressBarCounter()
     private var currentProgressStep = 0
     private var backPressedListener: BackPressedListener? = null
 
@@ -47,6 +47,18 @@ abstract class BaseWizardNavigator(
         val progressBar = mViewMvc.rootView?.findViewById<ProgressBar>(R.id.progress_bar)
         progressBar?.progress = currentProgressStep * STEP_PROGRESS
         progressBar?.max = progressBarCounter.currentProgressMax
+    }
+
+    open fun setupProgressBarMax(locationServicesAreOff: Boolean, areMapsDisabled: Boolean, isBluetoothDisabled: Boolean) {
+        if (locationServicesAreOff) {
+            progressBarCounter.increaseMaxProgress() // 1 additional step in flow
+        }
+        if (areMapsDisabled) {
+            progressBarCounter.increaseMaxProgress() // 1 additional step in flow
+        }
+        if (isBluetoothDisabled) {
+            progressBarCounter.increaseMaxProgress() // 1 additional step in flow
+        }
     }
 
     protected fun goToFragment(fragment: Fragment) {
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
@@ -16,7 +16,7 @@ abstract class BaseWizardNavigator(
     }
 
     protected abstract val STEP_PROGRESS: Int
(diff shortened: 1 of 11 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
```diff
@@ -15,6 +15,7 @@ abstract class BaseWizardNavigator(
     }
 
     protected abstract val STEP_PROGRESS: Int
+    open var MAX_PROGRESS = 40
```
mturnau, line 18: this is not really max progress because it can be higher ;) I'd name it `DEFAULT_MAX_PROGRESS` or similar

app/src/main/java/io/lunarlogic/aircasting/screens/new_session/NewSessionController.kt
```diff
@@ -80,6 +79,19 @@ class NewSessionController(
         }
     }
 
+    private fun setupProgressMax() {
+        if (!mContextActivity.areLocationServicesOn()) {
+            wizardNavigator.MAX_PROGRESS += 10
```
mturnau, line 84: typically we use capital letters for constants. Technically, it's not a const but var but when you use it like a var (increase) it should not be named with capital letters. You can have two things: a const to keep a default max and a var to keep the current max value

app/src/main/java/io/lunarlogic/aircasting/screens/common/BaseWizardNavigator.kt
```diff
@@ -15,6 +16,7 @@ abstract class BaseWizardNavigator(
     }
 
     protected abstract val STEP_PROGRESS: Int
+    val progressBarCounter = ProgressBarCounter(4) // we got 3 basic steps in the flow (progress bar should have 1 more to look better)
```
lesniakania, line 19: I'd move this `4` along with a comment and set it as a constant in `ProgressBarCounter` and use it in the constructor as default

app/src/main/java/io/lunarlogic/aircasting/screens/new_session/NewSessionController.kt
```diff
@@ -80,6 +76,19 @@ class NewSessionController(
         }
     }
 
+    private fun setupProgressMax() {
+        if (!mContextActivity.areLocationServicesOn()) {
+            wizardNavigator.progressBarCounter.increaseMaxProgress(1) // 1 additional step in flow
```
lesniakania, line 81: I'd make `increaseMaxProgress` method in the wizard navigator, use there `progressBarCounter.increaseMaxProgress` and set `1` as default

app/src/main/java/io/lunarlogic/aircasting/screens/settings/clear_sd_card/ClearSDCardController.kt
```diff
@@ -50,13 +50,28 @@ class ClearSDCardController(
...
+        setupProgressBarMax()
+
         if (mPermissionsManager.locationPermissionsGranted(mContextActivity)) {
             goToFirstStep()
         } else {
             mPermissionsManager.requestLocationPermissions(mContextActivity)
         }
     }
 
+    private fun setupProgressBarMax() {
```
lesniakania, line 62: and also seeing we would need to have the same things here as in `NewSessionController`, maybe we should move whole this method to the wizard class and pass `areLocationServicesOn` and other booleans as args?

lesniakania (changes requested):

app/src/main/java/io/lunarlogic/aircasting/screens/create_account/CreateAccountViewMvcImpl.kt
```diff
@@ -32,7 +32,6 @@ class CreateAccountViewMvcImpl : BaseObservableViewMvc<CreateAccountViewMvc.List
         val progressBarFrame = rootView?.findViewById<FrameLayout>(R.id.progress_bar_frame)
         if (!settings.onboardingDisplayed()) {
             progressBarFrame?.visibility = View.VISIBLE
-            settings.onboardingAccepted()
```
lesniakania, line 35: I guess this was temp thing ; ]?

PR #196, opened 2021-03-04
Fix deleting measurements

yzmopolis:
**Important changes:**

- I've added a function in order to get the last 1440 (24h) measurements from fixed sessions 
- I checked the date of the oldest one and passed this date to the measurements repository
- I've added deleting measurements from fixed sessions, but only older than this date

And it works for me. For testing how it works do it on your account and HabitatMap as well and:
- follow the below-listed sessions:
**4406, 766E, Office 2, Office**
-navigate to the fixed tab, pull down, and wait for the data to refresh

It should look good 🐱 if don't - try to reinstall the app as we established today during our morning standup 📴 

Screenshot 2021-03-04 at 17 50 29

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
@@ -50,11 +50,17 @@ interface MeasurementDao {
     @Query("DELETE FROM measurements")
     fun deleteAll()
 
-    @Query("DELETE FROM measurements WHERE time <= datetime('now', '-1 day') AND session_id in (:sessionIds)")
-    fun delete(sessionIds: List<Long>)
+    @Query("DELETE FROM measurements WHERE session_id in (:sessionIds) AND time < :lastExpectedMeasurementDate ")
+    fun delete(sessionIds: List<Long>, lastExpectedMeasurementDate: Date)
+
+    @Query("SELECT * FROM measurements WHERE session_id in (:sessionIds) ORDER BY time DESC LIMIT :limit")
+    fun getLastTwentyFourHours(sessionIds: List<Long>, limit: Int): List<MeasurementDBObject?>
+
+    @Query("SELECT * FROM measurements WHERE session_id in (:sessionIds) ORDER BY time DESC")
+    fun getAllMeasurementsBySessionIds(sessionIds: List<Long>): List<MeasurementDBObject?>
 
     @Transaction
-    fun deleteInTransaction(sessionIds: List<Long>) {
-        delete(sessionIds)
+    fun deleteInTransaction(sessionIds: List<Long>, lastExpectedMeasurementDate: Date) {
+        delete(sessionIds, lastExpectedMeasurementDate )
     }
 }
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
@@ -56,9 +56,6 @@ interface MeasurementDao {
     @Query("SELECT * FROM measurements WHERE session_id in (:sessionIds) ORDER BY time DESC LIMIT :limit")
     fun getLastTwentyFourHours(sessionIds: List<Long>, limit: Int): List<MeasurementDBObject?>
 
-    @Query("SELECT * FROM measurements WHERE session_id in (:sessionIds) ORDER BY time DESC")
-    fun getAllMeasurementsBySessionIds(sessionIds: List<Long>): List<MeasurementDBObject?>
-
     @Transaction
     fun deleteInTransaction(sessionIds: List<Long>, lastExpectedMeasurementDate: Date) {
         delete(sessionIds, lastExpectedMeasurementDate )
--- a/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
@@ -50,14 +50,14 @@ interface MeasurementDao {
     @Query("DELETE FROM measurements")
     fun deleteAll()
 
-    @Query("DELETE FROM measurements WHERE session_id in (:sessionIds) AND time < :lastExpectedMeasurementDate ")
-    fun delete(sessionIds: List<Long>, lastExpectedMeasurementDate: Date)
+    @Query("DELETE FROM measurements WHERE session_id=:sessionId AND time < :lastExpectedMeasurementDate ")
+    fun delete(sessionId: Long, lastExpectedMeasurementDate: Date)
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
```diff
@@ -50,11 +50,14 @@ interface MeasurementDao {
     @Query("DELETE FROM measurements")
     fun deleteAll()
 
-    @Query("DELETE FROM measurements WHERE time <= datetime('now', '-1 day') AND session_id in (:sessionIds)")
-    fun delete(sessionIds: List<Long>)
+    @Query("DELETE FROM measurements WHERE session_id in (:sessionIds) AND time < :lastExpectedMeasurementDate ")
+    fun delete(sessionIds: List<Long>, lastExpectedMeasurementDate: Date)
+
+    @Query("SELECT * FROM measurements WHERE session_id in (:sessionIds) ORDER BY time DESC LIMIT :limit")
+    fun getLastTwentyFourHours(sessionIds: List<Long>, limit: Int): List<MeasurementDBObject?>
```
mturnau, line 57: Actually, this function can get any X measurements and the name indicates it will always get 24h. You can either hardcode last 24h limit here or change the name so it is more generic

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -3,13 +3,24 @@ package io.lunarlogic.aircasting.networking.services
...
-        measurementRepository.deleteMeasurements(fixedSessionsIds)
+        val lastMeasurements = measurementRepository.getLastMeasurements(fixedSessionsIds, TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT)
+        if (checkIfDeleteMeasurements(lastMeasurements.size)) {
+            val lastExpectedMeasurement = lastMeasurements.last()
+            val lastExpectedMeasurementTime: Date = lastExpectedMeasurement?.time!!
+            measurementRepository.deleteMeasurements(fixedSessionsIds, lastExpectedMeasurementTime)
+        }
+    }
+
+    private fun checkIfDeleteMeasurements(measurementsCount: Int): Boolean {
```
mturnau, line 23: hm, maybe `shouldDeleteMeasurements`?

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -3,13 +3,24 @@ package io.lunarlogic.aircasting.networking.services
 import io.lunarlogic.aircasting.database.repositories.MeasurementsRepository
 import io.lunarlogic.aircasting.database.repositories.SessionsRepository
 import io.lunarlogic.aircasting.models.Session
+import java.util.*
 
 class RemoveOldMeasurementsService() {
     private val measurementRepository = MeasurementsRepository()
     private val sessionRepository = SessionsRepository()
+    private val TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT = 1440
```
mturnau, line 11: maybe we should put it as 24 * 60. Also, now we know what is it all about but it may become unclear in the future. The name of the class doesn't reflect that only fixed measurements are being deleted. Also, we could add a description comment that would explain why we delete 1440 measurements and that it doesn't necessarily mean we don't have last 24h data, rather 24h of data ;)

app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
yzmopolis: yeah, I forgot about it! good catch! :tada:

lesniakania: Also, I'm not sure if combining it for all sessions will work this way. I mean, I think it will return measurements for all fixed sessions and then limit it to 1440, so it will actually return much fewer measurements than needed.

I mean, I think e.g. for this scenario:

```
measurement_1_1 | session_1
... | ...
measurement_1_1440 | session_1
measurement_2_1 | session_2
... | ...
measurement_2_1440 | session_2
```

it will return only measurement_1_1...measurement_1_1440

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -3,13 +3,24 @@ package io.lunarlogic.aircasting.networking.services
...
 
     fun removeMeasurementsFromSessions() {
         val fixedSessionsIds = sessionRepository.sessionsIdsByType(Session.Type.FIXED)
-        measurementRepository.deleteMeasurements(fixedSessionsIds)
+        val lastMeasurements = measurementRepository.getLastMeasurements(fixedSessionsIds, TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT)
+        if (checkIfDeleteMeasurements(lastMeasurements.size)) {
+            val lastExpectedMeasurement = lastMeasurements.last()
+            val lastExpectedMeasurementTime: Date = lastExpectedMeasurement?.time!!
+            measurementRepository.deleteMeasurements(fixedSessionsIds, lastExpectedMeasurementTime)
```
lesniakania, line 19: I'd name it `deleteMeasurementsOlderThen` so it's more obvious what it does

app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -3,13 +3,24 @@ package io.lunarlogic.aircasting.networking.services
...
+        }
+    }
+
+    private fun checkIfDeleteMeasurements(measurementsCount: Int): Boolean {
+        return measurementsCount == TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT
```
lesniakania, line 24: I think it would be better to change it to `measurementsCount >= TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT`

yzmopolis: @lesniakania but it's not possible to have measurementsCount greater than `TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT` since this is the limit, am I wrong? 🤔

lesniakania: hm, ok, actually, you are right, because you explicitly set the limit, right

app/src/main/java/io/lunarlogic/aircasting/database/data_classes/measurements.kt
yzmopolis: @lesniakania you're totally right, i'll do it! 🚀

PR #197, opened 2021-03-05
Enhanced bottom sheets 🎉 🎉 🎉 

lesniakania:
https://trello.com/c/nS966waz/1152-enhance-bottom-sheets

Now we can release MVP 🤣 

I set default bottom sheet behavior to expanded and I fixed scrolling in share and delete bottom sheets.

Screenshot_20210305-091014_AircastingAndroid
Screenshot_20210305-090626_AircastingAndroid

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/common/BottomSheet.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/common/BottomSheet.kt
@@ -50,6 +50,7 @@ abstract class BottomSheet: BottomSheetDialogFragment() {
     }
 
     protected fun expandBottomSheet() {
+        // note that for this to work, you need to add bottomsheet_card id in the xml view
         val card = contentView?.bottomsheet_card ?: return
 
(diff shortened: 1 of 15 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/common/BottomSheet.kt
```diff
@@ -1,39 +1,61 @@
...
+        bottomSheetBehavior?.state = BottomSheetBehavior.STATE_EXPANDED
     }
 
     fun show(manager: FragmentManager) {
         show(manager, TAG)
     }
+
+    protected fun expandBottomSheet() {
+        // note that for this to work, you need to add bottomsheet_card id in the xml view
+        val card = contentView?.bottomsheet_card ?: return
```
mturnau, line 54: theoretically, we could pass this view as a parameter... not sure if it's worth it though..

lesniakania: @mturnau good point, I added optional argument - do you think it's ok?

PR #200, opened 2021-03-05
Increase radius corner of "Now" statistic box (because its bigger then "Avg" and "Peak" boxes)

michalrazny2:
https://trello.com/c/E0xnRSuh/1159-now-box-in-stats-over-the-map-graph-has-bigger-corner-radius-visually-because-its-bigger-than-peak-and-avg-decrease-radius-in-st

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
@@ -122,7 +122,7 @@ class StatisticsContainer {
         valueView?.text = Measurement.formatValue(value)
 
         val color = MeasurementColor.forMap(mContext, value, mSensorThreshold)
-        valueView?.background = StatisticsValueBackground(color, isNow)
+        valueView?.background = StatisticsValueBackground(color, StatisticsValueBackground.RADIUS_BIG)
         circleIndicator?.setColorFilter(color)
     }
 
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
@@ -102,7 +102,7 @@ class StatisticsContainer {
         if (mNow == null && stream != null) {
             mNow = getNowValue(stream)
         }
-        bindStatisticValues(stream, mNow, mNowValue, mNowCircleIndicator)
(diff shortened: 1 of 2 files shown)


lesniakania:
For me, it looks good, but I think @mturnau could look at this visually. I don't see such small differences so well.

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
```diff
@@ -102,7 +102,7 @@ class StatisticsContainer {
         if (mNow == null && stream != null) {
             mNow = getNowValue(stream)
         }
-        bindStatisticValues(stream, mNow, mNowValue, mNowCircleIndicator)
+        bindStatisticValues(stream, mNow, mNowValue, mNowCircleIndicator, true)
```
lesniakania, line 105: I'd change it to be less verbose that it's about "now", maybe rather allow to set radius here. You can leave the constant in the `StatisticsValueBackground `, but maybe let's name it more like `RADIUS_BIG` or something like that?

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/StatisticsContainer.kt
```diff
@@ -118,11 +118,11 @@ class StatisticsContainer {
         bindStatisticValues(stream, peak, mPeakValue, mPeakCircleIndicator)
     }
 
-    private fun bindStatisticValues(stream: MeasurementStream?, value: Double?, valueView: TextView?, circleIndicator: ImageView?) {
+    private fun bindStatisticValues(stream: MeasurementStream?, value: Double?, valueView: TextView?, circleIndicator: ImageView?, isNow: Boolean = false) {
         valueView?.text = Measurement.formatValue(value)
 
         val color = MeasurementColor.forMap(mContext, value, mSensorThreshold)
-        valueView?.background = StatisticsValueBackground(color)
+        valueView?.background = StatisticsValueBackground(color, StatisticsValueBackground.RADIUS_BIG)
```
lesniakania, line 125: Hm, I think now you increased the radius for all the boxes?
I'd change `isNow` in the method description to `radius: Float` and I'd pass `StatisticsValueBackground.RADIUS_BIG` when using in now stats:
`bindStatisticValues(stream, mNow, mNowValue, mNowCircleIndicator, StatisticsValueBackground.RADIUS_BIG)`

lesniakania:
I'm merging this, but @mturnau please check on the master how it looks in terms of radius value - is it ok now?

PR #203, opened 2021-03-08
Forgot password and create account links should be the same size

michalrazny2:
https://trello.com/c/GKqdojl2/1166-forgot-password-and-create-account-links-should-be-the-same-size


lesniakania:
I think it's fine, but I'd center those lines "Already have an account? Sign in" and "First time here? Create an account".

lesniakania:
Ok, we agreed to leave it as it is because it's according to the designs. We can revise it later on.

PR #210, opened 2021-03-09
Delete measurements by streams

yzmopolis:
Important changes: 

- I've added getting and deleting measurements for each stream in a fixed session and I've replaced previous logic with that.

In order to QA that it's good to reinstall the app when you switch branch to `delete-measurements-by-streams`. I compared data for few sessions in the old app and this branch, and it looks exactly the same. 

Screenshot 2021-03-10 at 12 35 16

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
@@ -21,7 +21,7 @@ class RemoveOldMeasurementsService() {
             if (shouldDeleteMeasurements(lastMeasurements.size)) {
                 val lastExpectedMeasurement = lastMeasurements.last()
                 val lastExpectedMeasurementTime: Date = lastExpectedMeasurement?.time!!
-                measurementRepository.deleteMeasurementsOlderThen(streamId, lastExpectedMeasurementTime)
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/lunarlogic/aircasting/networking/services/RemoveOldMeasurementsService.kt
```diff
@@ -2,22 +2,26 @@ package io.lunarlogic.aircasting.networking.services
...
-            val lastMeasurements = measurementRepository.getLastMeasurements(fixedSessionId, TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT)
+        val streamsForFixedSessionsIds = measurementStreamsRepository.getStreamsIdsBySessionIds(fixedSessionsIds)
+
+        streamsForFixedSessionsIds.forEach { streamId ->
+            val lastMeasurements = measurementRepository.getLastMeasurementsForStream(streamId, TWENTY_FOUR_HOURS_MEASUREMENTS_COUNT)
             if (shouldDeleteMeasurements(lastMeasurements.size)) {
                 val lastExpectedMeasurement = lastMeasurements.last()
                 val lastExpectedMeasurementTime: Date = lastExpectedMeasurement?.time!!
-                measurementRepository.deleteMeasurementsOlderThen(fixedSessionId, lastExpectedMeasurementTime)
+                measurementRepository.deleteMeasurementsOlderThen(streamId, lastExpectedMeasurementTime)
```
mturnau, line 24: Typo - should be OlderTh**a**n

PR #211, opened 2021-03-09
"Did you know image" update

michalrazny2:
https://trello.com/c/mObg6Km1/1167-did-you-know-box-hand-with-airbeam-should-have-transparent-gray-background-it-is-white-now

https://trello.com/c/BpT80T2X/1170-update-lets-start-screen-image-so-it-has-more-smooth-bottom-edge


mturnau:
IMO this image for onboarding background is still wrong. It has a grayish straight line ending and then there is white stripe with gradient. We need new image before we merge it.

PR #212, opened 2021-03-10
dashboard & graph & map: parameters are difficult to select via tapping

michalrazny2:
https://trello.com/c/vRCxVFTn/1088-dashboard-graph-map-parameters-are-difficult-to-select-via-tapping


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/MeasurementsTableContainer.kt
```diff
@@ -233,6 +233,12 @@ class MeasurementsTableContainer {
         }
 
         valueView.background = null
+        valueTextView.setOnClickListener {
+            onMeasurementClicked(stream)
+
+            markMeasurementHeaderAsSelected(stream)
+            markMeasurementValueAsSelected(stream)
```
mturnau, line 240: We use it in two places so it would be better if it was extracted as a function. Even slightly better (?), would be to pass the whole block to both bindStream and bindMeasurement methods in bindMeasurements() and use it in onclick listeners. Whatever seems more readable to you.

PR #213, opened 2021-03-11
Fixed bug outdoor sessions

yzmopolis:
I've added a new attribute to SessionPresenter: shouldHideMap - since sometimes sessionPresenter.session?.indoor didn't work properly. I'm still not sure why, but sometimes we have indoor == true in our DB and on the backend side as well, but the presenter "thought" there is false there.
Also, I've added an if-else statement in the session view. So far I didn't reproduce this bug but please double check it.
Screenshot 2021-03-14 at 20 05 45

How to QA it:

- Sign in as HabitatMap
- Go to fixed tab
- Follow a few sessions: 
"Michael Ave - Del Rey", "Queens College DEC Colocation", "Kalaheo - Kailua", "Office"
- Go to following tab
- Expand sessions and check if the map button is hidden for "Office" and shown for the rest
- Unfollow some of them and follow again, and check one more time
- If it works wait for a few minutes and check it again

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionPresenter.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionPresenter.kt
@@ -19,6 +19,7 @@ class SessionPresenter() {
     var sessionUUID: String? = null
     var initialSensorName: String? = null
     var visibleTimeSpan: ClosedRange<Date>? = null
+    var shouldHideMap: Boolean = false
 
     constructor(
         session: Session,
@@ -35,6 +36,7 @@ class SessionPresenter() {
         if (session.tab == SessionsTab.FOLLOWING || session.tab == SessionsTab.MOBILE_ACTIVE) {
             this.chartData = ChartData(session)
         }
+        this.shouldHideMap = session.indoor
 
         if (session.tab == SessionsTab.MOBILE_ACTIVE) {
             this.loading = true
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -204,8 +204,10 @@ abstract class SessionViewMvcImpl<ListenerType>: BaseObservableViewMvc<ListenerT
     }
 
     protected open fun bindMapButton(sessionPresenter: SessionPresenter) {
-        if (sessionPresenter.session?.indoor == true ) {
+        if (sessionPresenter.shouldHideMap) {
             mMapButton.visibility = View.GONE
+        } else {
+            mMapButton.visibility = View.VISIBLE
```
yzmopolis, line 210: it looks weird/redundant, but sometimes on loading/refresh it takes default from the presenter (which is false) and it's working only with if-else statement... <confused>

app/src/main/java/io/lunarlogic/aircasting/screens/dashboard/SessionPresenter.kt
```diff
@@ -35,6 +36,7 @@ class SessionPresenter() {
         if (session.tab == SessionsTab.FOLLOWING || session.tab == SessionsTab.MOBILE_ACTIVE) {
             this.chartData = ChartData(session)
         }
+        this.shouldHideMap = session.indoor
```
mturnau, line 39: Now that we know why it was not working sometimes, I think adding additional field is not necessary. The other change will be enough

PR #224, opened 2021-03-26
mobile active > "Did you know?" tip box: formatting should match the prototype

michalrazny2:
Screenshot 2021-03-26 at 12 34 28
https://trello.com/c/aH6lyN9D/1176-mobile-active-did-you-know-tip-box-formatting-should-match-the-prototype

Doubts:
- my phone does not seem to be representative to other android phones (we need to check these changes on other phones)
- 1. move AB+hand to bottom of box <- was not sure if I should really add some hard constraint so the image is at the bottom or just add margin instead of padding in the ImageView (I think it looks better and more like the proto)
- 2. added margins on left and right side of the image- not sure if its values are ok
- 3. line height of description text seems a bit to short but for some reason it did not want to change with "android:lineHeight" command on my phone

MARYSIA: 
Image looks fine (on the bottom)
left and right margins are ok
I added line height
I also updated the background color because in designs it had an opacity, which made it lighter


app/src/main/java/io/lunarlogic/aircasting/screens/main/MainController.kt
```diff
@@ -25,7 +25,7 @@ class MainController(
     private val mErrorHandler = ErrorHandler(rootActivity)
 
     fun onCreate() {
-        if (!mSettings.onboardingDisplayed() && mSettings.getAuthToken() == null) {  // todo: is this correct for sure <?>
+        if (!mSettings.onboardingDisplayed() && mSettings.getAuthToken() == null) {
```
michalrazny2, line 28: this is a random cleanup - just for you to know

PR #231, opened 2021-04-02
mobile active > map & graph: when loading long sessions, the spinner is missing

michalrazny2:
https://trello.com/b/UPSdV5fV/aircasting-android

Due to my research the bug happens when the map/graph are already loaded but the session is not downloaded yet. 
To this moment app was showing loader only during map loading without checking if the session is downloaded and possible to show. If the session wasn't downloaded the loader was hidden anyways and empty map appeared on screen.

I changed the condition so now app shows loader unless the list of measurement is not empty - it means the session is downloaded and ready to view.

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
@@ -92,13 +92,15 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
         mSessionPresenter = sessionPresenter
 
         bindSessionDetails()
-        if (sessionPresenter?.selectedStream != null) showSlider()
+        if (sessionPresenter?.selectedStream != null  && sessionPresenter.session == null) showSlider()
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
```diff
@@ -91,14 +91,15 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
...
 
-        bindSessionDetails()
-        if (sessionPresenter?.selectedStream != null) showSlider()
-
-        mMeasurementsTableContainer.bindSession(mSessionPresenter, this::onMeasurementStreamChanged)
-        bindStatisticsContainer()
-        mHLUSlider.bindSensorThreshold(sessionPresenter?.selectedSensorThreshold())
-
-        mSessionMeasurementsDescription?.visibility = View.VISIBLE
+        if (sessionPresenter?.selectedStream?.measurements?.isNotEmpty() == true) {
```
mturnau, line 94: Add TODO to check if it works properly after merging Marysia's changes ;)

PR #233, opened 2021-04-07
Memory leaks fixes and other improvements that make sense

mturnau:
Memory leak fixes:

- AppBar leaking (fixed by destroying class variable mTopAppBar on activity destroy)
- Fragments leaking - once fragment is destroyed, we should destroy View object so they don't keep reference to the activity. Achieved by adding BaseFragment and BaseController which do the right stuff

Other changes:

- NOT finishing active sessions unless the app crashed or was killed. If MainActivity was destroyed by Android system while app was in the background, we should keep the session recording
- Map:  pre-creating color spans that can be added to the polyline, one for each color level, Removing measurements from session presenter (not needed for displaying the map)
- Microphone: not saving 2 measurements per second


app/src/main/java/io/lunarlogic/aircasting/database/data_classes/sessions.kt
```diff
@@ -132,6 +132,9 @@ interface SessionDao {
     @Query("SELECT * FROM sessions WHERE uuid=:uuid AND deleted=0")
     fun loadLiveDataSessionAndMeasurementsByUUID(uuid: String): LiveData<SessionWithStreamsAndMeasurementsDBObject?>
 
+    @Query("SELECT * FROM sessions WHERE uuid=:uuid AND deleted=0")
+    fun reloadSessionAndStreamsByUUID(uuid: String): SessionWithStreamsDBObject?
```
mturnau, line 136: remove this and the rest of removing observer

PR #234, opened 2021-04-08
When I create two microphone sessions only one gets measurements

michalrazny2:
https://trello.com/c/xs4BVFhV/882-when-i-create-two-microphone-sessions-only-one-gets-measurements


app/src/main/java/io/lunarlogic/aircasting/screens/new_session/NewSessionController.kt
```diff
@@ -142,6 +141,21 @@ class NewSessionController(
...
 
     override fun onMicrophoneDeviceSelected() {
+        GlobalScope.launch(Dispatchers.Main) {
+            var existing = false
+            val query = GlobalScope.async(Dispatchers.IO) {
+                existing = sessionsRepository.isMicrophoneSessionAlreadyRecording()
+            }
+            query.await()
+            if (existing) {
+                errorHandler.showError(mContextActivity.getString(R.string.you_cant_start_2_microphone_sessions_at_once))
```
mturnau, line 151: indentation got wrong I think

PR #235, opened 2021-04-13
[OLD APPROACH] Avg and peak not calculated properly

michalrazny2:
https://trello.com/c/Z8nXozLq/1144-following-graph-avg-and-peak-not-calculated-properly-after-zooming-panning


mturnau:
Random comment: let's check if we can do anything with disabling "super fast drag" gesture - I mean, there is some "velocity" setting in fling gesture and I wonder if setting it/reading it could help us

PR #236, opened 2021-04-13
following & fixed: graph doesn't immediately update

michalrazny2:
https://trello.com/c/TnSGmXAu/1199-following-fixed-graph-doesnt-immediately-update


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/graph/GraphContainer.kt
```diff
@@ -86,7 +86,7 @@ class GraphContainer: OnChartGestureListener {
         drawData(entries)
         drawMidnightPointLines(result.midnightPoints)
         drawThresholds()
-        setLabels()
+        //setLabels()
```
michalrazny2, line 90: I'm not sure why this line sets wrong labels... but it seems that commenting this out gives us correct labels. This function might be not necessary considering that we call "drawLabels" function in zoom() function (5 lines earlier)

mturnau (approved):

PR #242, opened 2021-04-16
dashboard: "Delete session" doesn't always delete the session

michalrazny2:
https://trello.com/c/N7R2jQY5/1171-dashboard-delete-session-doesnt-always-delete-the-session

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/networking/services/PeriodicallySyncSessionsService.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/networking/services/PeriodicallySyncSessionsService.kt
@@ -1,0 +1,63 @@
+package io.lunarlogic.aircasting.networking.services
+
+import io.lunarlogic.aircasting.exceptions.ErrorHandler
+import io.lunarlogic.aircasting.lib.Settings
+import io.lunarlogic.aircasting.networking.responses.SyncResponse
+import retrofit2.Call
+
+class PeriodicallySyncSessionsService(
+    private val apiService: ApiService,
+    private val errorHandler: ErrorHandler,
+    private val settings: Settings,
+    private val sessionsSyncService: SessionsSyncService
+    ) {
+    private val thread = SyncThread()
+
+    fun start() {
+        thread.start()
+    }
+
+    fun stop() {
+        thread.cancel()
+    }
+
+    fun pause() {
+        thread.paused = true
+    }
+
+    fun resume() {
+        thread.paused = false
+    }
+
+    inner class SyncThread(): Thread() {
+        private val POLL_INTERVAL = 60 * 1000L // 1 minute
+        var paused = false
+        private var call: Call<SyncResponse>? = null
+
+
+        override fun run() {
+            try {
+                while (!isInterrupted && settings.getAreThereSessionsToRemove() == true) {
+                    syncSessions()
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
```diff
@@ -219,8 +219,11 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
 
     private fun deleteSession(sessionUUID: String) {
         DatabaseProvider.runQuery {
+            settings.setDeletingSessionsInProgress(true)
             sessionsRespository.markForRemoval(listOf(sessionUUID))
             sessionsSyncService.sync()
+//            settings.setSessionsToRemove(true) todo: is this needed for sure??
+            settings.setDeletingSessionsInProgress(false)
         }
```
michalrazny2, line 227: @mturnau Due to your advice I added a variable to check if "Delete session is in progress"- I use it to prevent starting another sync() when one is already in progress. Beyond that, I was supposed to add another variable to check if there are some sessions waiting to be removed- for now I abandoned this idea as I thought maybe "deletingSessionInProgress" variable might be enough to fix deleting sessions. I am not sure about possible consequences of this approach (maybe I forgot about something?).

mturnau: The part you did will work well to prevent syncing during deletion. However, after you finish deleting sessions, you have no guarantee the sync will be triggered. I suggested adding a service checking periodically if there are any sessions to delete and perform sync if so.

app/src/main/java/io/lunarlogic/aircasting/networking/services/PeriodicallySyncSessionsService.kt
```diff
@@ -0,0 +1,63 @@
+package io.lunarlogic.aircasting.networking.services
+
+import io.lunarlogic.aircasting.exceptions.ErrorHandler
+import io.lunarlogic.aircasting.lib.Settings
+import io.lunarlogic.aircasting.networking.responses.SyncResponse
+import retrofit2.Call
+
+class PeriodicallySyncSessionsService(
```
mturnau, line 8: I think we should name it "PeriodicallySyncRemovedSessionsService" to make it visible it will only happened when we remove sessions

mturnau: An afterthought: all of the sessions will be synced if you remove sessions. So maybe let's not change the name but add a prominent comment expaining what is id doing and that it will only sync sessions if there are any deleted pending.

app/src/main/java/io/lunarlogic/aircasting/sensor/SessionManager.kt
```diff
@@ -219,8 +224,11 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
 
     private fun deleteSession(sessionUUID: String) {
         DatabaseProvider.runQuery {
+            settings.setDeletingSessionsInProgress(true)
             sessionsRespository.markForRemoval(listOf(sessionUUID))
             sessionsSyncService.sync()
```
mturnau, line 229: I guess this sync here is unnecessary, it will never perform, right?

PR #243, opened 2021-04-21
Displaying notes- new branch

michalrazny2:


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -329,13 +328,6 @@ class MapContainer: OnMapReadyCallback {
 
         return mapOptions
     }
-
-    private fun clearPresenterMeasurements() {
-        mSessionPresenter?.session?.streams?.forEach {
-            it.clearMeasurements()
-        }
-        mSessionPresenter?.selectedStream?.clearMeasurements()
```
mturnau, line 307: I think you should also delete the method in MeasurementStream

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/graph/GraphDataGenerator.kt
```diff
@@ -40,6 +48,19 @@ class GraphDataGenerator {
 
                 entries.add(buildAverageEntry(date))
 
+                if (notes != null) {
+                    for (note in notes) {
+                        if (isSameDate(note, date) && note != previousNote) { // TODO: check if this condition is still needed after fixing bug with 2 measurements in 1 second
```
mturnau, line 53: this is now fixed on master so you can check it

app/src/main/java/io/lunarlogic/aircasting/models/Session.kt
```diff
@@ -332,4 +337,8 @@ class Session(
         formatter.timeZone = TimeZone.getDefault()
         return formatter
     }
+
+    fun lastMeasurementLocation(): LatLng {
+        return LatLng(streams.first().measurements.last().latitude!!, streams.first().measurements.last().longitude!!)
```
mturnau, line 342: I don't like this method. First code smell is many dots - looks like Session should not know so much about where is lat and long of last measurement. Session may get its first stream, then stream may know where is its last measurement. I refactored it and will show you the commit.

Second thing is about double bang. It will throw NullPointerException if lat or long are null. And they may be null, becuase they are nullable types. I handled this case with logging and displaying error.

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
```diff
@@ -153,6 +154,13 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
         mListener?.onSensorThresholdChanged(sensorThreshold)
     }
 
+    protected open fun onNoteAdded(note: Note) {
+        val oldNotes = mSessionPresenter?.session?.notes
+        if (oldNotes != null) {
+            mSessionPresenter?.session?.notes = oldNotes + note
```
mturnau, line 160: maybe it would make sense to use MutableList for notes instead of this logic? If you are in fact changing this list, there is no point in pretending it's immutable

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/graph/GraphContainer.kt
```diff
@@ -37,23 +38,27 @@ class GraphContainer: OnChartGestureListener {
...
     private val DATE_FORMAT = "HH:mm"
     private val mDefaultZoomSpan: Int?
     private var shouldZoomToDefault = true
     private var mOnTimeSpanChanged: (timeSpan: ClosedRange<Date>) -> Unit
     private var mGetMeasurementsSample: () -> List<Measurement>
     private var mMeasurementsSample: List<Measurement> = listOf()
+    private var mNotes: List<Note>? = listOf()
 
-    constructor(rootView: View?, context: Context, defaultZoomSpan: Int?, onTimeSpanChanged: (timeSpan: ClosedRange<Date>) -> Unit, getMeasurementsSample: () -> List<Measurement>) {
+    constructor(rootView: View?, context: Context, defaultZoomSpan: Int?, onTimeSpanChanged: (timeSpan: ClosedRange<Date>) -> Unit, getMeasurementsSample: () -> List<Measurement>, getNoteSample: List<Note>?) {
```
mturnau, line 51: name "get" suggests it's a function name and it is not - it's a variable. Name "sample" suggests that you only have a sample of notes here restricted by time and it is not true - you pass all of the notes. It may be named just "notes"

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/graph/GraphController.kt
```diff
@@ -48,4 +50,16 @@ class GraphController(
         val event = NoteCreatedEvent(session, note)
         EventBus.getDefault().post(event)
     }
+
+    override fun noteMarkerClicked(session: Session?, noteNumber: Int) {
```
mturnau, line 54: When I tested it, nothing happened when I clicked the note on the graph - not sure if it was intended

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -32,9 +34,11 @@ class MapContainer: OnMapReadyCallback {
     private val mLocateButton: ImageView?
     private var mMapFragment: SupportMapFragment?
     private var mSupportFragmentManager: FragmentManager?
+    private val mMarkers: HashMap<String?, Int?> = hashMapOf() // TODO: Clear this variable onDestroy to prevent memory leaks
```
mturnau, line 37: now you can do it in destroy() method

PR #248, opened 2021-04-28
Map: map buttons

michalrazny2:
https://trello.com/c/8yuED33O/1204-map-map-buttons


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -102,6 +102,12 @@ class MapContainer: OnMapReadyCallback {
         clearMap()
 
         mMap?.isBuildingsEnabled = false
+        mMap?.uiSettings?.isZoomControlsEnabled = false
```
mturnau, line 105: this should be set in mapOptions()

PR #260, opened 2021-05-22
[WIP] Storing authentication token in more secure way

michalrazny2:
https://trello.com/c/POzjqnbD/1040-store-user-authentication-token-in-settings-in-the-more-secure-way


LinX64:
Same here - will check it later.

PR #264, opened 2021-05-25
mobile dormant > share file: can't share file

michalrazny2:
https://trello.com/c/4aP9vs8P/1230-mobile-dormant-share-file-cant-share-file


app/src/main/java/io/lunarlogic/aircasting/networking/services/ExportSessionService.kt
```diff
@@ -17,7 +18,7 @@ class ExportSessionService(private val apiService: ApiService, private val error
                 if (response.isSuccessful) {
                     successCallback.invoke()
                 } else {
-                    errorHandler.handle(SessionExportFailedError())
+                    errorHandler.handle(SessionUploadPendingError())
```
mturnau, line 21: in sessionsController in shareLinkPressed we could display the same error so we don't have pne copy in two places
`mErrorHandler.handleAndDisplay(SessionUploadPendingError())` (?)

PR #267, opened 2021-05-27
[WIP] Chart - new way of generating data

mturnau:


mturnau:
Closing for now as we won't be needing this in near future

PR #273, opened 2021-06-02
mobile dormant > graph & map: don't show stats container

michalrazny2:
https://trello.com/c/NhkdVP9j/1245-mobile-dormant-graph-map-dont-show-stats-container

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
@@ -90,7 +90,7 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
     override fun bindSession(sessionPresenter: SessionPresenter?) {
         mSessionPresenter = sessionPresenter
 
-        if (sessionPresenter?.session?.tab == SessionsTab.MOBILE_DORMANT) mStatisticsContainer = null
+        bindStatisticsContainer()
         if (sessionPresenter?.selectedStream?.measurements?.isNotEmpty() == true) {
             bindSessionDetails()
             showSlider()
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
```diff
@@ -89,13 +90,15 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
     override fun bindSession(sessionPresenter: SessionPresenter?) {
         mSessionPresenter = sessionPresenter
 
+        if (sessionPresenter?.session?.tab == SessionsTab.MOBILE_DORMANT) mStatisticsContainer = null
```
mturnau, line 93: We should not need this line. There is a method bindStatisticsContainer() which is overriden in MapViewMobileDormantMvcImpl and it sets stats container to null as well. Try to check what is going on there because it should just work.

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/SessionDetailsViewMvcImpl.kt
```diff
@@ -96,13 +102,18 @@ abstract class SessionDetailsViewMvcImpl: BaseObservableViewMvc<SessionDetailsVi
...
-            bindStatisticsContainer()
 
+            bindStatisticsContainer()
             mHLUSlider.bindSensorThreshold(sessionPresenter?.selectedSensorThreshold())
             mSessionMeasurementsDescription?.visibility = View.VISIBLE
         }
     }
 
+    protected fun shouldShowStatisticsContainer(): Boolean {
+        //todo: do nadpisania w dormancie żeby zwracała false
```
michalrazny2, line 113: this is not done yet I guess? or not pushed from local branch at least

PR #274, opened 2021-06-02
mobile dormant > map: don't display map dot

michalrazny2:
https://trello.com/c/rBmR9eON/1224-mobile-dormant-map-dont-display-map-dot

### DIFF
--- a/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
+++ b/app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
@@ -198,9 +198,9 @@ class MapContainer: OnMapReadyCallback {
 
 
     private fun drawLastMeasurementMarker(point: LatLng?, color: Int?) {
+        if (!shouldDrawLastMeasurement()) return
         if (point == null || color == null) return
         if (mLastMeasurementMarker != null) mLastMeasurementMarker!!.remove()
-        if (!shouldDrawLastMeasurement()) return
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -162,7 +163,7 @@ class MapContainer: OnMapReadyCallback {
         mMeasurementsLineOptions.addAll(mMeasurementPoints)
         mMeasurementsLine = mMap?.addPolyline(mMeasurementsLineOptions)
 
-        if (latestPoint != null && latestColor != null) {
+        if (latestPoint != null && latestColor != null  && mSessionPresenter?.session?.tab != SessionsTab.MOBILE_DORMANT) {
```
mturnau, line 166: It would be better to create a function returning Boolean shouldDrawLastMeasurement or sth and check it at the beginning of drawLastMeasurementMarker function, return if false.

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -199,6 +200,7 @@ class MapContainer: OnMapReadyCallback {
     private fun drawLastMeasurementMarker(point: LatLng?, color: Int?) {
         if (point == null || color == null) return
         if (mLastMeasurementMarker != null) mLastMeasurementMarker!!.remove()
+        if (!shouldDrawLastMeasurement()) return
```
mturnau, line 203: I'd move it as a top if

PR #281, opened 2021-06-21
Settings > backend settings: After changing backend url some requests are still using the previous URL

michalrazny2:
https://trello.com/c/roNI9hGj/1269-settings-backend-settings-after-changing-backend-url-some-requests-are-still-using-the-previous-url


mturnau:
Well ok, it still doesn't work ;-) I mean partially... but not on the jobs: PeriodicallyDownloadFixedSessionMeasurementsService and probably others from sessionManager too. Not sure how to fix it yet. They are created in SessionManager which persists...

PR #284, opened 2021-06-24
mobile active > map: when entering the map, map should be centered over current location

michalrazny2:
https://trello.com/c/Vbffqm7B/1258-mobile-active-map-when-entering-the-map-map-should-be-centered-over-current-location

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
+++ b/app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.content.res.Resources
 import android.graphics.Color
 import android.location.Location
+import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
@@ -29,6 +30,7 @@ import java.util.concurrent.atomic.AtomicInteger
 
 class MapContainer: OnMapReadyCallback {
     private val DEFAULT_ZOOM = 16f
+    private var currentZoom : Float? = null
 
     private var mContext: Context?
     private var mListener: SessionDetailsViewMvc.Listener? = null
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -242,6 +242,13 @@ class MapContainer: OnMapReadyCallback {
         val boundingBox = SessionBoundingBox.get(mMeasurements)
         val padding = 100 // meters
         mMap?.animateCamera(CameraUpdateFactory.newLatLngBounds(boundingBox, padding))
+
+        if (mSessionPresenter?.isRecording() == true) {
+            mMap?.moveCamera(CameraUpdateFactory.newLatLngZoom(LatLng(mMeasurements.last().latitude!!, mMeasurements.last().longitude!!),
+                50F // 50 is hardcoded zoom value for now
```
mturnau, line 248: leaving this PR for now because we need to think how to avoid hardcoding zoom level. I'm not sure if it will look ok for all of the sessions

app/src/main/java/io/lunarlogic/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -242,6 +242,13 @@ class MapContainer: OnMapReadyCallback {
         val boundingBox = SessionBoundingBox.get(mMeasurements)
         val padding = 100 // meters
         mMap?.animateCamera(CameraUpdateFactory.newLatLngBounds(boundingBox, padding))
+
+        if (mSessionPresenter?.isRecording() == true) {
+            mMap?.moveCamera(CameraUpdateFactory.newLatLngZoom(LatLng(mMeasurements.last().latitude!!, mMeasurements.last().longitude!!),
```
mturnau, line 247: We can try using the zoom that is used after animating to session
`val currentZoom = mMap?.cameraPosition?.zoom`

mturnau: Let's communicate to Michael that in some cases (when the last measurement is on the verge) not the whole session will be visible. If he always wants to see the whole session, we'd have to calculate the visible bounds differently, which is possible but more complicated. Let's get feedback on a simple solution.

mturnau: `centerMap(LatLng(mMeasurements.last().latitude!!, mMeasurements.last().longitude!!))`

PR #310, opened 2021-08-12
Settings: use 24 hour format

michalrazny2:
https://trello.com/c/nSM9nS4g/1082-settings-use-24-hr-format

TODO:
- how can I retrieve settings field (using24HourFormat) to use it on View classes ??
- add HOUR_FORMAT logic to GraphContainer


app/src/main/java/pl/llp/aircasting/lib/DateConverter.kt
```diff
@@ -6,7 +6,25 @@ import java.util.*
...
+        private val DATE_FORMAT_24 = "HH:mm"
+        private val DATE_FORMAT_12 = "hh:mm a"
+
+        fun setup(settings: Settings) {
+            if (singleton == null) DateConverter()
+            mSettings = settings
+        }
+
+        fun createToDateString(date: Date, timeZone: TimeZone = TimeZone.getDefault(), dateFormat: String = DEFAULT_DATE_FORMAT): String {
+            if (mSettings?.isUsing24HourFormat() == true) {
```
mturnau, line 22: you can use some class variable that you set in setup, e.g.
```
mDateFormat = if (mSettings?.isUsing24HourFormat() == true)
DATE_FORMAT_24
else 
DATE_FORMAT_12
```

and then just call `toDateString(date, timeZone, mDateFormat)`

app/src/main/java/pl/llp/aircasting/lib/DateConverter.kt
```diff
@@ -6,7 +6,25 @@ import java.util.*
...
         val DEFAULT_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+        private val DATE_FORMAT_24 = "HH:mm"
+        private val DATE_FORMAT_12 = "hh:mm a"
+
+        fun setup(settings: Settings) {
+            if (singleton == null) DateConverter()
+            mSettings = settings
+        }
+
+        fun createToDateString(date: Date, timeZone: TimeZone = TimeZone.getDefault(), dateFormat: String = DEFAULT_DATE_FORMAT): String {
```
mturnau, line 21: We should have two methods, one fore.g. params, with default format and one for display, which should always use appropriate hour format (12 or 24)

app/src/main/java/pl/llp/aircasting/lib/DateHelper.kt
```diff
@@ -0,0 +1,76 @@
...
+            } else {
+                return durationString12HourFormat(session)
+            }
+        }
+
+        fun durationString(session: Session): String {
+            val dateFormatter = dateTimeFormatter(DATE_FORMAT)
+            val hourFormatter = dateTimeFormatter(HOUR_FORMAT_24)
+
+            var durationString = "${dateFormatter.format(session.startTime)} ${hourFormatter.format(session.startTime)}"
```
mturnau, line 35: DateConverter. toTimeStringForDisplay(session.startTime)
DateCoverter.toDateStrongForDisplay(session.endTime)

app/src/main/java/pl/llp/aircasting/lib/DateConverter.kt
```diff
@@ -1,12 +1,22 @@
...
+        val singleton: DateConverter? = null
+        var mSettings: Settings? = null
+
         val DEFAULT_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss"
+        private val DATE_FORMAT = "MM/dd/yy"
+        private val HOUR_FORMAT_24 = "HH:mm"
+        private val HOUR_FORMAT_12 = "hh:mm a"
+
+        fun setup(settings: Settings) {
+            if (singleton == null) DateConverter()
```
mturnau, line 17: singleton is always null. You never set it. What you do here is not a singleton pattern

mturnau: I thibk you can remove singleton var all together and it still should work?

app/src/main/java/pl/llp/aircasting/lib/DurationStringHelper.kt
```diff
@@ -0,0 +1,20 @@
+package pl.llp.aircasting.lib
+
+import java.util.*
+
+class DurationStringHelper {
+        fun durationString(startTime: Date, endTime: Date?): String {
+            var durationString = "${DateConverter.get()?.toDateStringForDisplay(startTime)} ${DateConverter.get()?.toTimeStringForDisplay(startTime)}"
```
mturnau, line 7: assign DateConverter object to a val

app/src/main/java/pl/llp/aircasting/lib/DateConverter.kt
```diff
@@ -1,27 +1,62 @@
...
-        fun toDateString(
-            date: Date,
-            timeZone: TimeZone = TimeZone.getDefault(),
-            dateFormat: String = DEFAULT_DATE_FORMAT
-        ): String {
-            val formatter = SimpleDateFormat(dateFormat, Locale.getDefault())
-            formatter.timeZone = timeZone
-            return formatter.format(date)
+        fun get(): DateConverter {
+            return singleton!!
```
mturnau, line 15: I don't like the idea of double bang here. It will only not throw null pointer exception if you fire setup() method first and it's not guaranteed. It's better you return an optional and handle it in the place you use the get() method.

app/src/main/java/pl/llp/aircasting/lib/DateConverter.kt
```diff
@@ -1,27 +1,62 @@
...
+        return formatter.format(date)
+    }
+
+    fun isTheSameDay(startTime: Date, endTime: Date): Boolean {
+        val dateFormat = SimpleDateFormat("yyyyMMdd")
+        return dateFormat.format(startTime) == dateFormat.format(endTime)
+    }
+
+    fun toTimeStringForDisplay(date: Date, timeZone: TimeZone = TimeZone.getDefault()): String {
+        if (singleton?.mSettings?.isUsing24HourFormat() == true) {
```
mturnau, line 51: you rather should not have to call it through singleton, it's a class variable?

mturnau: also, I'd prefer
`val hourFormat = if (mSettings?.isUsing24HourFormat() == true)
HOUR_FORMAT_24
else 
HOUR_FORMAT_12`
and
`return toDateString(date, timeZone, hourFormat)`

app/src/main/java/pl/llp/aircasting/networking/params/SessionParams.kt
```diff
@@ -19,8 +19,8 @@ class SessionParams {
 
         this.contribute = session.contribute
         this.title = session.name
-        this.start_time = DateConverter.toDateString(session.startTime)
-        this.end_time = DateConverter.toDateString(session.endTime!!)
+        this.start_time = DateConverter.get().toDateString(session.startTime)
```
mturnau, line 22: assign DateConverter.get() to a val

PR #322, opened 2021-08-27
mobile active & mobile dormant > 2+ hrs > delete note: can't delete note, note is deleted from the app map & graph UI but not from the web

michalrazny2:
https://trello.com/c/28NA6gJb/1322-mobile-active-mobile-dormant-2-hrs-delete-note-cant-delete-note-note-is-deleted-from-the-app-map-graph-ui-but-not-from-the-web

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/sensor/SessionManager.kt
+++ b/app/src/main/java/pl/llp/aircasting/sensor/SessionManager.kt
@@ -309,15 +309,6 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
     }
 
     private fun updateSession(session: Session) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/pl/llp/aircasting/sensor/SessionManager.kt
```diff
@@ -317,6 +317,16 @@ class SessionManager(private val mContext: Context, private val apiService: ApiS
         }
     }
 
+    private fun updateSessionWithNotes(session: Session) {
+        val reloadedSession = sessionsRespository.loadSessionForUpload(session.uuid)
```
mturnau, line 321: check if you can just use loadSessionForUpload in updateSession method instead of creating a separate one

PR #326, opened 2021-09-09
Target api level 30

michalrazny2:
TODO:
[ ] more testing on older and newer Android devices <?>
[ ] maybe: removing warnings given by kotlin compiler


app/src/main/java/pl/llp/aircasting/exceptions/ErrorHandler.kt
```diff
@@ -23,7 +23,7 @@ class ErrorHandler(private val mContext: Context): Handler(Looper.getMainLooper(
 
     fun handle(exception: BaseException) {
         exception.cause?.printStackTrace()
-        exception.messageToDisplay?.let { Log.e(TAG, exception.messageToDisplay) }
+        exception.messageToDisplay?.let { message -> if (exception.messageToDisplay is String) Log.e(TAG, message) }
```
michalrazny2, line 26: I changed this line according to this thread:
https://stackoverflow.com/questions/41086296/smartcast-is-impossible-because-property-has-open-or-custom-getter

app/src/main/java/pl/llp/aircasting/sensor/AirBeamClearCardService.kt
```diff
@@ -37,7 +37,7 @@ class AirBeamClearCardService: AirBeamService() {
     override fun startSensor(intent: Intent?) {
         intent ?: return
 
-        val deviceItem = intent.getParcelableExtra(AirBeamRecordSessionService.DEVICE_ITEM_KEY) as DeviceItem
+        val deviceItem = DeviceItem(intent.getParcelableExtra(AirBeamRecordSessionService.DEVICE_ITEM_KEY))
```
michalrazny2, line 40: Kotlin does not accept "as DeviceItem" cast because we got custom getter/setter:
https://stackoverflow.com/questions/41086296/smartcast-is-impossible-because-property-has-open-or-custom-getter (first answer)
After changing it to DeviceItem() syntax it compiled properly

PR #328, opened 2021-09-10
Android 7 sync

mturnau:
Allowing Android 7 and below devices to sync data from SD card


mturnau:
I tried to fix this for Android 7 by lowering MTU size but in the end it didn't work for Micheal so we decdide it to ditch it. However, the solution might be explored firther if we ever need it.

PR #343, opened 2021-10-22
Settings: keep screen on

michalrazny2:
https://trello.com/c/FRWFkBu9/1081-settings-keep-screen-on

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -13,8 +13,8 @@ android {
         applicationId "pl.llp.aircasting"
         minSdkVersion 21
         targetSdkVersion 30
-        versionCode 150
-        versionName "2.0.25"
+        versionCode 153
+        versionName "2.0.26"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 
         // Read the API key from ./secure.properties into R.string.maps_api_key
--- a/app/src/main/java/pl/llp/aircasting/lib/Settings.kt
+++ b/app/src/main/java/pl/llp/aircasting/lib/Settings.kt
@@ -16,7 +16,6 @@ open class Settings(private val mApplication: Application) {
     protected val MAPS_DISABLED_KEY = "maps_disabled"
     protected val BACKEND_URL_KEY = "backend_url"
     protected val BACKEND_PORT_KEY = "backend_port"
-    protected val AIRBEAM3_CONNECTED_KEY = "airbeam3_connected" // this flag is used to check if airbeam3 was connected to the phone in the past
     protected val ONBOARDING_DISPLAYED_KEY = "onboarding_displayed"
     protected val APP_RESTARTED = "app_restarted"
     protected val KEEP_SCREEN_ON_KEY = "keep_screen_on"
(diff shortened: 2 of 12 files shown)


app/src/main/java/pl/llp/aircasting/screens/new_session/NewSessionController.kt
```diff
@@ -90,7 +91,12 @@ class NewSessionController(
         wizardNavigator.setupProgressBarMax(!mContextActivity.areLocationServicesOn(), settings.areMapsDisabled(), !bluetoothManager.isBluetoothEnabled())
     }
 
+    fun onResume() {
+        if (settings.isKeepScreenOnEnabled()) mContextActivity.window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+    }
+
     fun onStop() {
+        mContextActivity.window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
```
mturnau, line 99: why? won't this clear flag when you leave this controller?

app/src/main/java/pl/llp/aircasting/screens/sync/SyncActivity.kt
```diff
@@ -75,6 +75,11 @@ class SyncActivity: AppCompatActivity() {
         AppBar.setup(view.rootView, this)
     }
 
+    override fun onResume() {
+        super.onResume()
+        controller?.onResume()
```
mturnau, line 80: could we, instead of adding this to every controller, create base activity with onResume method overriden with this thing?

mturnau: Anyway, it's not even in all of the activities. Wouldn't the screen go off if you are e.g. on graph activity?

michalrazny2: According to second comment: I think that screen shouldn't go off on Graph/Map activities as we set this flag on SessionDetailsViewController which is super class for both views' controller

michalrazny2: According to first comment: I suppose that this base activity would be superclass for all activities in our app and we don't want to add this behaviour to all activities (e.g. LoginActivity, CreateAccountActivity). In this case we would have to override onResume() method from BaseActivity in these Activities with an empty method... or something like this ?? I'm not sure how to approach this one.

app/src/main/java/pl/llp/aircasting/screens/new_session/NewSessionController.kt
michalrazny2: Hmm.. when I was coding it I had in mind such case:  user sets "Keep screen on"= true -> user enters NewSessionActivity (we set flags to keep screen on) -> user sets "Keep screen on"=false -> in such case we will never clear this flag for our NewSessionActivity. However as I think about I it now, every time we destroy one instance of NewSessionActivity (after we start recording new session) and create new one (when we want to start recording another session) we set this flag everytime for a "fresh" instance of the activity, so I guess you are right :)

mturnau: I would suppose the flag is gone anyway when the activity is destroyed... but we can double check and test

app/src/main/java/pl/llp/aircasting/screens/sync/SyncActivity.kt
mturnau: what we could do: create a method in base activity e.g. shouldKeepScreenOnIfSet and set it default, to true (or false, depending on which situation is more often) and then override it in those we don't /dowant to use it, then in base onResume go with 
if (shouldKeepScreenOnIfSet() //i don;t really love this method name) {
// set the flag on activity
}
Now I'm thinking that if you reset all of the settings on logout (keepscreenon also?), then having this in Logn and Create account won't harm us (every time you'd be on these screns, keep screen on should be set to false...)

PR #344, opened 2021-10-25
let's begin > text formatting: reformat button text so it's 2 lines rather than 3

michalrazny2:
https://trello.com/c/iyAsgpUY/1339-lets-begin-text-formatting-reformat-button-text-so-its-2-lines-rather-than-3


app/src/main/java/pl/llp/aircasting/screens/lets_start/LetsStartViewMvcImpl.kt
```diff
@@ -41,16 +41,11 @@ class LetsStartViewMvcImpl: BaseObservableViewMvc<LetsStartViewMvc.Listener>,
         val orLabel = rootView?.or
         val syncCard = rootView?.sync_card
 
-        if (mSettings.airbeam3Connected()) {
```
mturnau, line 44: did you remove this if? you should not I guess ;) testing purposes?

mturnau:
Completely not sure if this has anything to do here but mind that we do have different font sizes for smaller screens height under 600 (you'll find the values in values/dimens.xml and those for bigger screens will be in values-h600dp/dimens). So when testing this, try on virstual devices (or real, if you have) with really small screen)

PR #348, opened 2021-11-30
Airbeam not paired connection failed

michalrazny2:
https://trello.com/c/izAfoVeP/1358-config-wizard-ab2-thats-never-been-paired-prior-bt-connection-fails


app/src/main/java/pl/llp/aircasting/sensor/AirBeamConnector.kt
```diff
@@ -20,7 +20,7 @@ abstract class AirBeamConnector {
     }
 
     private var mListener: Listener? = null
-    private val CONNECTION_TIMEOUT = 10000L
+    private val CONNECTION_TIMEOUT = 20000L
```
mturnau, line 23: are we sure we want to change this?

PR #350, opened 2021-12-01
Following tab

mturnau:


app/src/main/java/pl/llp/aircasting/screens/dashboard/following/FollowingController.kt
```diff
@@ -23,7 +24,7 @@ class FollowingController(
 ): SessionsController(mRootActivity, mViewMvc, mSessionsViewModel, mSettings, mApiServiceFactory, mRootActivity!!.supportFragmentManager, mContext),
     SessionsViewMvc.Listener {
 
-    private var mSessionsObserver = ActiveSessionsObserver(mLifecycleOwner, mSessionsViewModel, mViewMvc)
+    private var mSessionsObserver = MobileActiveSessionsObserver(mLifecycleOwner, mSessionsViewModel, mViewMvc)
```
mturnau, line 27: we should change the name of this observer if we want to use it on following tab. Maybe just ActiveSessionsObserver because i don't think we use current ActvieSessionObserver anywhere else

app/src/main/java/pl/llp/aircasting/networking/services/DownloadMeasurementsCallback.kt
```diff
@@ -90,6 +92,16 @@ class DownloadMeasurementsCallback(
...
         measurementsRepository.insertAll(streamId, sessionId, measurements)
+        if (session.isFixed()) {
+            // #TODO we do not need to pass activeSessionMeasurementsRepository for mobiel sessions, i should be an optional argument
+            measurements.forEach { measurement ->
+                activeSessionMeasurementsRepository.createOrReplace(
+                    sessionId,
+                    streamId,
+                    measurement
+                )
+            }
```
mturnau, line 103: To do:

- [ ] implement bulk inserts for cases when we have more than one measurement to insert (e.g. when we log in). We need to replace as many measurements as we get (max 540 tho)
- [ ] perform this only for followed sessions
- [ ] when you follow a session, you should populate this table using measurements table last 540 measurmeents
- [ ] when you unfollow a session, you should delete all the records from this table for given session

app/src/main/java/pl/llp/aircasting/database/data_classes/active_sessions_measurements.kt
```diff
@@ -45,9 +45,15 @@ interface ActiveSessionMeasurementDao {
     @Query("SELECT id FROM active_sessions_measurements WHERE session_id=:sessionId AND stream_id=:streamId ORDER BY time ASC LIMIT 1")
     fun getOldestMeasurementId(sessionId: Long, streamId: Long): Int
 
+    @Query("SELECT id FROM active_sessions_measurements WHERE session_id=:sessionId AND stream_id=:streamId ORDER BY time ASC LIMIT :limit")
+    fun getOldestMeasurementsId(sessionId: Long, streamId: Long, limit: Int): List<Int>
```
mturnau, line 49: if this can return multiple ids, the method name should be ids

app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
```diff
@@ -49,4 +69,44 @@ class ActiveSessionMeasurementsRepository {
...
+                    ActiveSessionMeasurementDBObject(
+                        measurementStreamId,
+                        sessionId,
+                        measurement.value,
+                        measurement.time,
+                        measurement.latitude,
+                        measurement.longitude
+                    )
+                )
+            }
```
mturnau, line 105: few lines above are repeted also in 78-87. We should DRY this out

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -213,6 +213,9 @@ class MeasurementStream(
     }
 
     fun lastMeasurement(): Measurement {
+        if (measurements.isNotEmpty())
         return measurements.last()
+        else
+            return Measurement()
```
mturnau, line 219: not sure why we need this.. I mean, default measurement doesn't make much sense to me because it is not a meaningful data

app/src/main/java/pl/llp/aircasting/networking/services/DownloadMeasurementsCallback.kt
```diff
@@ -90,6 +92,9 @@ class DownloadMeasurementsCallback(
             Measurement(response, averagingFrequency)
         }
         measurementsRepository.insertAll(streamId, sessionId, measurements)
+        if (session.isFixed() && session.followedAt != null) {
+            activeSessionMeasurementsRepository.createOrReplaceMultipleRows(streamId, sessionId, measurements)
```
mturnau, line 96: this is one of the places where some comment explaining what are we doing would be good

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -255,4 +264,50 @@ abstract class SessionsController(
...
+        return measurements.mapNotNull { measurementDBObject ->
+            measurementDBObject?.let { measurement ->
+                Measurement(measurement)
+            }
+        }
+    }
+
+    private fun addFollowedSessionMeasurementsToActiveTable(session: Session) {
+        DatabaseProvider.runQuery {
+            val sessionId = mSessionRepository.getSessionIdByUUID(session.uuid)
```
mturnau, line 278: this logic (copying some measurements to active table) should not be in controller, all should be in db related class

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -255,4 +264,50 @@ abstract class SessionsController(
...
+            sessionId?.let { loadMeasurementsForStreams(it, session.streams) }
+        }
+    }
+
+    private fun clearUnfollowedSessionMeasurementsFromActiveTable(session: Session) {
+        DatabaseProvider.runQuery {
+            val sessionId = mSessionRepository.getSessionIdByUUID(session.uuid)
```
mturnau, line 285: this as well

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -255,4 +264,50 @@ abstract class SessionsController(
...
+            mActiveSessionsRepository.deleteBySessionId(sessionId)
+        }
+    }
+
+    private fun loadMeasurementsForStreams(
```
mturnau, line 290: see above

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsRecyclerAdapter.kt
```diff
@@ -46,7 +46,7 @@ abstract class SessionsRecyclerAdapter<ListenerType>(
         removeObsoleteSessions()
         sessions.forEach { session ->
             if (mSessionPresenters.containsKey(session.uuid)) {
-                val sessionPresenter = mSessionPresenters[session.uuid]
+                val sessionPresenter = mSessionPresenters[session.uuid] //todo: czy w tym ifie mamy napewno najnowsze measurementy dla naszej following sesji
```
mturnau, line 49: we should have the newest as this is from the observer. However, if it's active table that's being observed, there may be the case when we haven't copied it from the main table yet?

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
michalrazny2: I'm not sure why, but the app kept crashing when we were calling measurements.last() on empty list at the application start

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
michalrazny2: hmm... okey, to be honest I wanted to do it the same way it was done in SessionDetailsViewController and right now I'm a bit confused where should I place the db-connected code from SessionsController, I guess we can discuss it tomorrow

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
mturnau: then we should make it optional and check if it's not null

PR #351, opened 2021-12-13
Auto reconnect - fixes for AB3 & AB2

mturnau:


app/src/main/java/pl/llp/aircasting/sensor/AirBeamReconnector.kt
```diff
@@ -155,6 +156,7 @@ class AirBeamReconnector(
     }
 
     private fun finalizeReconnection() {
+        mAirBeamDiscoveryService.reset()
```
mturnau, line 159: we were not nulling Discovery service and it qwould use the same callbacks with next discovery, hence "auto-reconnect is triggered while on the "Choose the device you'd like to record with" screen. this premature attempt to connect to the AB2 when the connection indicator is still red creates a situation where the config wizard continues past the "Choose the device you'd like to record with" screen without user prompting while the AB2 connection indicator is still solid red. when you complete the wizard, the AB2 and the app never connect properly."

app/src/main/java/pl/llp/aircasting/sensor/airbeam2/AirBeam2Connector.kt
```diff
@@ -75,11 +75,11 @@ open class AirBeam2Connector(
                     mAirBeam2Reader.run(socket.inputStream)
                 }
             } catch(e: IOException) {
-                mErrorHandler.handle(SensorDisconnectedError("called from Airbeam2Connector: IOException"))
+                mErrorHandler.handle(SensorDisconnectedError("called from Airbeam2Connector: IOException connection established ${connectionEstablished.get()}"))
                 onDisconnected(deviceItem)
-                onConnectionFailed(deviceItem)
 
                 if (!cancelStarted.get() && !connectionEstablished.get()) {
+                    onConnectionFailed(deviceItem)
```
mturnau, line 82: this fixed reconnect for AB2. This callback was triggered after successful reconnect and it was triggering reconnect again

app/src/main/java/pl/llp/aircasting/sensor/airbeam3/AirBeam3Connector.kt
```diff
@@ -84,14 +80,19 @@ open class AirBeam3Connector(
...
-    override fun onDeviceDisconnecting(device: BluetoothDevice) {}
+    override fun onDeviceDisconnecting(device: BluetoothDevice) {
+        mErrorHandler.handle(SensorDisconnectedError("called from Airbeam3Connector onDeviceDisconnecting"))
+    }
     override fun onDeviceDisconnected(device: BluetoothDevice, reason: Int) {
         val deviceItem = DeviceItem(device)
+
         mErrorHandler.handle(SensorDisconnectedError("called from Airbeam3Connector onDeviceDisconnected device id ${deviceItem.id} reason ${reason}"))
         onDisconnected(deviceItem)
+        disconnect()
```
mturnau, line 96: this in fact is doing BluetoothGatt.close() which is needed to avoid double sending disconnected event

PR #352, opened 2021-12-13
mobile dormant > map & graph > avg value: inaccurate avg values when stream is not selected

michalrazny2:
https://trello.com/c/kRk6o0WI/1374-mobile-dormant-map-graph-avg-value-inaccurate-avg-values-when-stream-is-not-selected

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/screens/session_view/SessionDetailsViewController.kt
+++ b/app/src/main/java/pl/llp/aircasting/screens/session_view/SessionDetailsViewController.kt
@@ -212,12 +212,13 @@ abstract class SessionDetailsViewController(
             val streamId =
                 MeasurementStreamsRepository().getId(sessionId, measurementStream)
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/pl/llp/aircasting/screens/session_view/SessionDetailsViewController.kt
```diff
@@ -212,9 +212,10 @@ abstract class SessionDetailsViewController(
             val streamId =
                 MeasurementStreamsRepository().getId(sessionId, measurementStream)
 
+            val session = mSessionRepository.getSessionById(sessionId)
```
mturnau, line 215: At the point you call this method (loadMeasurementsForStreams) you have session object. Makes little sense to get it again from DB. You could check if session is mobile dormant (?) before calling this method and pass it as a boolean parameter.

PR #366, opened 2022-03-01
Map Locate fix with zero values

LinX64:
This pull request is for the issue with the map locate where the values were zeros and it couldn't find the right location for it.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
+++ b/app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import android.content.res.Resources
 import android.graphics.Color
 import android.location.Location
-import android.util.Log
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
@@ -13,6 +12,7 @@ import androidx.core.content.res.ResourcesCompat
 import androidx.fragment.app.FragmentManager
 import com.google.android.libraries.maps.*
 import com.google.android.libraries.maps.model.*
+import kotlinx.android.synthetic.main.activity_map.view.*
 import pl.llp.aircasting.R
 import pl.llp.aircasting.lib.BitmapHelper
 import pl.llp.aircasting.lib.MeasurementColor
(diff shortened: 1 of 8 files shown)


LinX64:
I created this `map-locate-fix` branch from the master with old libraries, I think we will need to check the other pull request for updates + code refactoring first.

app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -330,10 +337,26 @@ class MapContainer: OnMapReadyCallback {
     }
 
     private fun locate() {
-        if (mSessionPresenter?.isMobileActive() == true) {
-            mListener?.locateRequested()
+        val latitude = mSessionPresenter?.session?.location?.latitude
+        val longitude = mSessionPresenter?.session?.location?.longitude
+
+        if (latitude != 0.0 || longitude != 0.0) {
+
```
mturnau, line 344: why these extra empty lines?

app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -330,10 +337,26 @@ class MapContainer: OnMapReadyCallback {
...
+                mSessionPresenter?.session?.location
+            )
+
         } else {
-            centerMap(mSessionPresenter?.session?.location)
+
+            var lati: Double? = 0.0
+            var long: Double? = 0.0
+
+            for (measurement in mMeasurements) {
```
mturnau, line 354: can't you just get last measurement location straight away instead of iterating through all of them?

app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
```diff
@@ -330,10 +337,26 @@ class MapContainer: OnMapReadyCallback {
...
+                lati = measurement.latitude
```
mturnau, line 355: we use rather `lat` for latitude throughout the app

app/src/main/java/pl/llp/aircasting/screens/session_view/map/MapContainer.kt
LinX64: Good point, done!

mturnau (approved):

PR #382, opened 2022-03-14
Chart wrong averages

xameYeoh:
This one does not have chart averages reversed as it was before. The averages are still a bit different from the one in iOS

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
+++ b/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
@@ -8,7 +8,7 @@ import pl.llp.aircasting.models.Measurement
 import pl.llp.aircasting.models.MeasurementStream
 
 class ActiveSessionMeasurementsRepository {
-    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 10 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
     private val mDatabase = DatabaseProvider.get()
 
     fun insert(measurementStreamId: Long, sessionId: Long, measurement: Measurement): Long {
--- a/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
+++ b/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
@@ -8,7 +8,7 @@ import pl.llp.aircasting.models.Measurement
 import pl.llp.aircasting.models.MeasurementStream
 
 class ActiveSessionMeasurementsRepository {
-    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 10 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
     private val mDatabase = DatabaseProvider.get()
 
     fun insert(measurementStreamId: Long, sessionId: Long, measurement: Measurement): Long {
--- a/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
+++ b/app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
@@ -1,6 +1,5 @@
 package pl.llp.aircasting.database.repositories
 
-import androidx.sqlite.db.SimpleSQLiteQuery
 import pl.llp.aircasting.database.DatabaseProvider
 import pl.llp.aircasting.database.data_classes.ActiveSessionMeasurementDBObject
 import pl.llp.aircasting.database.data_classes.MeasurementDBObject
@@ -8,7 +7,10 @@ import pl.llp.aircasting.models.Measurement
 import pl.llp.aircasting.models.MeasurementStream
 
 class ActiveSessionMeasurementsRepository {
-    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    companion object {
+        const val MAX_MEASUREMENTS_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    }
+
     private val mDatabase = DatabaseProvider.get()
 
     fun insert(measurementStreamId: Long, sessionId: Long, measurement: Measurement): Long {
@@ -50,7 +52,7 @@ class ActiveSessionMeasurementsRepository {
     fun createOrReplace(sessionId: Long, streamId: Long, measurement: Measurement) {
         val lastMeasurementsCount = mDatabase.activeSessionsMeasurements().countBySessionAndStream(sessionId, streamId)
 
-        if (lastMeasurementsCount > ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER ) {
+        if (lastMeasurementsCount > MAX_MEASUREMENTS_NUMBER ) {
             val activeSessionMeasurementDBObject = ActiveSessionMeasurementDBObject(
                 streamId,
                 sessionId,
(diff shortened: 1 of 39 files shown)


app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
```diff
@@ -8,7 +8,7 @@ import pl.llp.aircasting.models.Measurement
 import pl.llp.aircasting.models.MeasurementStream
 
 class ActiveSessionMeasurementsRepository {
-    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 10 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
```
mturnau, line 11: please update comment so it reflects the reality (we only need 10 mins/hours...)

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -69,6 +68,7 @@ class MeasurementStream(
         this.mMeasurements = streamWithLastMeasurementsDBObject.measurements.map { measurementDBObject ->
             Measurement(measurementDBObject)
         }
+        this.mMeasurements = this.mMeasurements.sortedWith(compareByDescending { it.time })
```
mturnau, line 71: I think this doesn't have to be in a separate line? you can sort right after maping?

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -169,32 +169,31 @@ class MeasurementStream(
...
         // copy the backing list to avoid ConcurrentModificationException
-        val allMeasurements = ArrayList<Measurement>(measurements)
-
+        val allMeasurements = ArrayList(measurements)
         val measurementsSize = allMeasurements.size
 
         if (amount >= measurementsSize) return allMeasurements
 
-        return allMeasurements.subList(measurementsSize - amount, measurementsSize)
+        return allMeasurements.subList(0, amount)
```
mturnau, line 179: if I get it correctly, after this change the method will return first X measurements, not last... if you always start from 0 index. We have to be careful cause we use the same method to get measurements for graph

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -212,7 +211,7 @@ class MeasurementStream(
         return filteredMeasurements.takeLast(amount)
     }
 
-    fun lastMeasurement(): Measurement? {
+    fun lastMeasurement(): Measurement {
```
mturnau, line 214: how come this can NOT be optional? what if there are no measurements in the stream? it will return null

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -40,6 +40,7 @@ abstract class SessionsController(
     val fragmentManager: FragmentManager,
     private var context: Context?
 ) : SessionsViewMvc.Listener, EditSessionBottomSheet.Listener, ShareSessionBottomSheet.Listener, DeleteSessionBottomSheet.Listener {
+    private val MAX_CHART_MEASUREMENTS_NEEDED = 600
```
mturnau, line 43: I don't like that name much. MAX_CHART_MEASUREMENTS_NUMBER or something would be better

app/src/main/java/pl/llp/aircasting/screens/dashboard/ShareSessionBottomSheet.kt
```diff
@@ -129,7 +130,7 @@ class ShareSessionBottomSheet(
         radioButton.id = View.generateViewId()
         radioButton.text = stream.detailedType
         radioButton.layoutParams = layoutParams
-        radioButton.setTextAppearance(context, R.style.TextAppearance_Aircasting_Body1)
+        radioButton.setAppearance(context!!, R.style.TextAppearance_Aircasting_Body1)
```
mturnau, line 133: not sure if double banging context is a good idea?

app/src/main/java/pl/llp/aircasting/screens/dashboard/charts/ChartAveragesCreator.kt
```diff
@@ -85,18 +85,16 @@ class ChartAveragesCreator {
...
+    fun getFixedEntries(stream: MeasurementStream): MutableList<Entry> {
+        val measurements: MutableList<Measurement>?
         var xValue = MAX_X_VALUE.toDouble()
         val entries: MutableList<Entry> = mutableListOf()
         val periodData: MutableList<List<Measurement>?> = mutableListOf()
-        val maxMeasurementsAmount = 600
 
-        measurements = stream.getLastMeasurements(maxMeasurementsAmount)
+        measurements = stream.getLastMeasurements()
+        removeMeasurementsFromUnfinishedHour(measurements)
```
mturnau, line 95: not sure, but if you just remove measurements from unfinished hour, you will still have too many measurements from the beginning.
if it's 11:15, you got 10h of measurements which is 1:15-11:15. You remove 11:00-11:15 and you're left with 1:15-11 and what we really want is 2:00-11:00. Am I correct?

app/src/main/java/pl/llp/aircasting/screens/dashboard/charts/ChartAveragesCreator.kt
```diff
@@ -106,33 +104,49 @@ class ChartAveragesCreator {
...
+        calendar.time = measurements.first().time
+
+        val unwantedHour = calendar[Calendar.HOUR_OF_DAY]
+
+        while (true) {
+            calendar.time = measurements.first().time
+            val lastMeasurementHour = calendar[Calendar.HOUR_OF_DAY]
+            if (lastMeasurementHour == unwantedHour) {
+                measurements.removeFirst()
+            } else break
```
mturnau, line 148: I have a feeling we do as simple thing in a complicated way. Maybe there is a way to use filter() or sth?

app/src/main/java/pl/llp/aircasting/screens/dashboard/charts/ChartAveragesCreator.kt
mturnau: Also, I reckon we could add some comment exmplaining this. In few months we'll have no idea what it all was about.

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -169,14 +169,20 @@ class MeasurementStream(
...
         val allMeasurements = ArrayList(measurements)
         val measurementsSize = allMeasurements.size
 
         if (amount >= measurementsSize) return allMeasurements
 
-        return allMeasurements.subList(0, amount)
+        return allMeasurements.subList(measurementsSize - amount, measurementsSize)
+    }
+
+    fun getFreshestMeasurementValue(): Double {
```
mturnau, line 182: getLastMeasurementValue would be more consistent name

mturnau: also, could be
`lastMeasurement()?.value ?: 0.0`

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -211,7 +217,7 @@ class MeasurementStream(
         return filteredMeasurements.takeLast(amount)
     }
 
-    fun lastMeasurement(): Measurement {
+    fun lastMeasurement(): Measurement? {
         return measurements.last()
```
mturnau, line 221: now when I'm looking at it I'm not sure but we may need lastOrNull(). Wouldn't last throw an exception?

app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
```diff
@@ -1,14 +1,16 @@
...
 import pl.llp.aircasting.database.DatabaseProvider
 import pl.llp.aircasting.database.data_classes.ActiveSessionMeasurementDBObject
 import pl.llp.aircasting.database.data_classes.MeasurementDBObject
 import pl.llp.aircasting.models.Measurement
 import pl.llp.aircasting.models.MeasurementStream
 
 class ActiveSessionMeasurementsRepository {
-    private val ACTIVE_SESSIONS_MEASUREMENTS_MAX_NUMBER = 60 * 10 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
+    companion object {
+        const val MAX_MEASUREMENTS_NUMBER = 60 * 9 // we only need 9 mins of measurements. TODO: we should calculate that number based on time
```
mturnau, line 11: ee.. we're back to 540?

app/src/main/java/pl/llp/aircasting/models/MeasurementStream.kt
```diff
@@ -169,32 +169,35 @@ class MeasurementStream(
         return measurements.filter { it.time in timeSpan}
     }
 
-    fun getLastMeasurements(amount: Int): MutableList<Measurement>? {
+    fun getFreshMeasurements(amount: Int = ActiveSessionMeasurementsRepository.MAX_MEASUREMENTS_NUMBER): MutableList<Measurement> {
```
mturnau, line 172: why fresh not last? it takes amount as a parameter, so it makes sense as it returns you N last measurements. "Fresh" is not as informative IMO

app/src/main/java/pl/llp/aircasting/screens/dashboard/SessionsController.kt
```diff
@@ -40,6 +35,7 @@ abstract class SessionsController(
     val fragmentManager: FragmentManager,
     private var context: Context?
 ) : SessionsViewMvc.Listener, EditSessionBottomSheet.Listener, ShareSessionBottomSheet.Listener, DeleteSessionBottomSheet.Listener {
+    private val MAX_CHART_MEASUREMENTS_NUMBER = 600
```
mturnau, line 38: why it's 600 here and 540 there? If we keep only 540 in that table, we'll never get 600, right? actually, we could use the same const in two places..

PR #383, opened 2022-03-15
Reoccurring toast

LinX64:
I believe this happened after one hour because of updating the measurements and getting fresh data from server. Also, I disabled the background refresh for the time the app is paused and user is not using the application, but in the resume situation, while user is using the app, the app will do its work.

This is also optimized in newest Android APIs which will ask the user to see if user allows the app to run in background or not.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/networking/services/SessionsSyncService.kt
+++ b/app/src/main/java/pl/llp/aircasting/networking/services/SessionsSyncService.kt
@@ -213,6 +213,6 @@ class SessionsSyncService {
     }
 
     fun pause() {
-        syncInBackground.set(false)
+        syncInBackground.set(true)
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/pl/llp/aircasting/networking/services/SessionsSyncService.kt
```diff
@@ -214,6 +213,6 @@ class SessionsSyncService {
     }
 
     fun pause() {
-        syncInBackground.set(true)
+        syncInBackground.set(false)
```
mturnau, line 216: this change is not disabling sync in background. This is preventing to sync after the app goes to foreground. If app is n background and sync is triggered, in line 86 after this change it will return false and continue sync, which is exactly what we DON'T want. 
I don't know where is the cause if the problem but I'd rather check if onAppToBackground() is called properly from SessionManager when we go to background, if it is the sync must be triggered somewhere else cause pausing sync used to work well. Maybe this is somehow connected to the updates (like we don't catch lifecycle events well now? - check AppLifecycleObserver, it looks deprecated now, maybe we need to change it)

LinX64: It was also interesting to me at first, seems like it's because of the deprecated AppLifecycleObserver. Thanks for mentioning it.

PR #388, opened 2022-03-21
Active fixed no chart measurements

xameYeoh:
This requires some testing. I have tested it by some shenanigans making chart to draw fixed session measures every minute instead of hourly and it appeared to be working


app/src/main/java/pl/llp/aircasting/database/repositories/ActiveSessionMeasurementsRepository.kt
```diff
@@ -75,17 +75,12 @@ class ActiveSessionMeasurementsRepository {
     }
 
     fun createOrReplaceMultipleRows(measurementStreamId: Long, sessionId: Long, measurements: List<Measurement>) {
-        val lastMeasurementsCount = mDatabase.activeSessionsMeasurements().countBySessionAndStream(sessionId, measurementStreamId)
-        var measurementsToBeReplaced = listOf<Measurement>()
+        var measurementsToBeReplaced = measurements
 
         if(measurements.size > MAX_MEASUREMENTS_NUMBER) {
```
mturnau, line 80: maybe we should add here some comment explaining what is it all about

PR #389, opened 2022-03-23
Active fixed no chart measurements

xameYeoh:


app/src/main/java/pl/llp/aircasting/screens/dashboard/charts/ChartData.kt
```diff
@@ -25,6 +25,7 @@ class ChartData(
     private var mChartRefreshService = ChartRefreshService(session)
 
     init {
+        // TODO: See if the ChartData instantiates only when following and unfollowing or at some other time as well
```
mturnau, line 28: this actuall happens also when we have mobile avtive session and we want to display chart

PR #400, opened 2022-03-28
Navcontroller Memory Leak Fix

LinX64:
This is the second part of the memory leaks issue for NavigationController.


LinX64:
It may need a few improvements as we tested it, I'll create another PR to separate everything and not to confuse people.

PR #414, opened 2022-04-05
Airbeam Reconnect Issue | Plan B

LinX64:
This is the implementation for Airbeam reconnect issue. The user should go to settings - battery optimization to be able to use the foreground service while running in the background. 

Also, this PR includes a new extension function for checking API versions.


app/src/main/java/pl/llp/aircasting/screens/new_session/session_details/FixedSessionDetailsController.kt
```diff
@@ -90,11 +90,11 @@ class FixedSessionDetailsController(
             scanForNetworks()
         }
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+        if (isSDKVersionBiggerThanQ()) {
             val intent = Intent(Settings.Panel.ACTION_WIFI)
             startForResult?.launch(intent)
         } else {
-            mWifiManager?.setWifiEnabled(true)
```
xameYeoh, line 97: As far as I got it, the method is deprecated after Q Android version. In else block we have scenario where we have Android version less than Q, so we can keep the method? I mean now it's just a getter for wifiState, or am I missing something?

app/src/main/java/pl/llp/aircasting/screens/session_view/graph/GraphContainer.kt
```diff
@@ -17,13 +17,17 @@ import com.github.mikephil.charting.listener.ChartTouchListener
 import com.github.mikephil.charting.listener.OnChartGestureListener
 import com.github.mikephil.charting.listener.OnChartValueSelectedListener
 import kotlinx.android.synthetic.main.graph.view.*
+<<<<<<< airbeam_reconnect_issue_plan_b
+=======
 import kotlinx.coroutines.DelicateCoroutinesApi
+>>>>>>> master
```
xameYeoh, line 23: Conflicts?

app/src/main/java/pl/llp/aircasting/screens/new_session/session_details/FixedSessionDetailsController.kt
```diff
@@ -90,11 +90,11 @@ class FixedSessionDetailsController(
             scanForNetworks()
         }
 
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+        if (isSDKVersionBiggerThanQ()) {
```
xameYeoh, line 93: I think in all the extensions like this it would be better to have naming like "Greater**OrEqual**" to avoid confusion.

xameYeoh (commented):
Comments

app/src/main/java/pl/llp/aircasting/screens/new_session/session_details/FixedSessionDetailsController.kt
LinX64: Good catch, I didn't notice that.

app/src/main/java/pl/llp/aircasting/screens/new_session/session_details/FixedSessionDetailsController.kt
```diff
@@ -94,7 +94,7 @@ class FixedSessionDetailsController(
             val intent = Intent(Settings.Panel.ACTION_WIFI)
             startForResult?.launch(intent)
         } else {
-            mWifiManager?.wifiState
```
xameYeoh: It was [setWifiEnabled()](https://developer.android.com/reference/android/net/wifi/WifiManager#setWifiEnabled(boolean)), not isWifiEnabled(), right?

xameYeoh (commented):
Method replacement

app/src/main/java/pl/llp/aircasting/screens/new_session/session_details/FixedSessionDetailsController.kt
```diff
@@ -94,7 +94,7 @@ class FixedSessionDetailsController(
...
+            mWifiManager?.isWifiEnabled = true
```
xameYeoh, line 97: It was setWifiEnabled(true), not isWifiEnabled, right?

LinX64: Actually, both are the same. In Kotlin, we can use it like this to set the value to `true`. Check [this link](https://developer.android.com/reference/android/net/wifi/WifiManager#isWifiEnabled()) and this image: https://imgur.com/a/dDNj5z4 Let's say it was the old method.

PR #416, opened 2022-04-06
Graph lines zig-zagging after SD sync

xameYeoh:
Measurements used in graph were losing correct sort order after sync which resulted in stray lines (zig-zagging of one line)


LinX64 (approved):
Looks good. Let's test.

PR #417, opened 2022-04-08
Duplicate measurements in db | Less measurements in chart

xameYeoh:
The issue with graph having less measurements for some sessions was due to duplicates in the DB for these sessions. We already had a onConflict REPLACEMENT policy for duplicate rows, but we did not have Index that would tell SQLite what rows to consider duplicates. I've added a unique Columns combination to tell which rows are considered duplicates - ones that have same session_id, stream_id and time altogether


LinX64 (approved):
Looks good, let's test now.

PR #418, opened 2022-04-11
Spinner disappears before all sessions have loaded

LinX64:


app/src/main/java/pl/llp/aircasting/networking/services/PeriodicallySyncSessionsService.kt
```diff
@@ -36,7 +36,7 @@ class PeriodicallySyncSessionsService(
 
         override fun run() {
             try {
-                while (!isInterrupted && settings.getAreThereSessionsToRemove() == true) {
+                while (!isInterrupted && settings.getAreThereSessionsToRemove()) {
```
xameYeoh, line 39: This naming seems strange. Should we rename the method to settings.thereAreSessionsToRemove() ?

xameYeoh (approved):

PR #421, opened 2022-04-14
Adding follow session card lets begin screen

xameYeoh:


LinX64:
Merged in the huge PR.

PR #422, opened 2022-04-14
Search fixed sessions screen

xameYeoh:


LinX64:
Same here, this PR is also merged in the huge PR.

PR #424, opened 2022-04-15
SEARCH FEATURE - Search map (without bottom cards)

LinX64:
This is the PR for the search feature which includes adding icon, creating map screen and etc.


LinX64:
Already merged in https://github.com/HabitatMap/AircastingAndroid/pull/432. No need for this.

PR #425, opened 2022-04-16
AirBeamSyncService:  lateinit property mContext has not been initialized

LinX64:
This is a fix for context memory leak issue on DatabaseProvider and also improvments for the project.


LinX64:
[AirBeamSyncService: lateinit property mContext has not been initialized](https://trello.com/c/56gvEbGG/1481-airbeamsyncservice-lateinit-property-mcontext-has-not-been-initialized)

app/src/main/java/pl/llp/aircasting/database/DatabaseProvider.kt
```diff
@@ -39,18 +38,14 @@ abstract class AppDatabase : RoomDatabase() {
     abstract fun activeSessionsMeasurements(): ActiveSessionMeasurementDao
 }
 
-class DatabaseProvider {
+abstract class DatabaseProvider {
     companion object {
-        private val DB_NAME = "aircasting"
-
-        private lateinit var mContext: Context
```
xameYeoh, line 46: Can we create inject the application context here using Dagger?

app/src/main/java/pl/llp/aircasting/database/DatabaseProvider.kt
```diff
@@ -39,18 +38,14 @@ abstract class AppDatabase : RoomDatabase() {
...
+        private const val DB_NAME = "aircasting"
 
-        fun setup(context: Context) {
-            mContext = context
-        }
+        @Volatile
+        private var mAppDatabase: AppDatabase? = null
 
-        fun get(): AppDatabase {
+        fun get(mContext: Context): AppDatabase {
```
xameYeoh, line 48: Injecting the Application context would eliminate the need to constantly have it as a parameter here (we really mean Application here, not a Context, right? As we create DB instance only once for the whole App)

app/src/main/java/pl/llp/aircasting/AircastingApplication.kt
```diff
@@ -34,4 +39,8 @@ class AircastingApplication: Application() {
             .addObserver(AppLifecycleObserver())
 
     }
+
+    companion object {
+        lateinit var appContext: Context
```
xameYeoh, line 44: Injecting would eliminate the need for storing the AppContext in this static field as well

app/src/main/java/pl/llp/aircasting/database/DatabaseProvider.kt
```diff
@@ -39,18 +38,14 @@ abstract class AppDatabase : RoomDatabase() {
     abstract fun activeSessionsMeasurements(): ActiveSessionMeasurementDao
 }
 
-class DatabaseProvider {
+abstract class DatabaseProvider {
```
xameYeoh, line 41: Do we have any classes that extend DatabaseProvider? If we want to say that we do not need to instantiate the class, we could have a private constructor instead of marking the class abstract

xameYeoh (commented):
I think that we could have an injection of application context in DatabaseProvider, instead of constantly providing it as a parameter. This would provide lazy initialisation on its own as well

PR #426, opened 2022-04-17
Fatal Exception: android.database.sqlite.SQLiteException

LinX64:
This is a fix for the Fatal exception bug.


LinX64:
[Fatal Exception: android.database.sqlite.SQLiteException](https://trello.com/c/B1oZW0DD/1483-fatal-exception-androiddatabasesqlitesqliteexception)

xameYeoh:
This was already present in my [pr](https://github.com/HabitatMap/AircastingAndroid/pull/417/commits/a76a454b709e384a888571fa2ea142b9a7dc67b6)
image

app/src/main/java/pl/llp/aircasting/database/DatabaseProvider.kt
```diff
@@ -18,7 +18,7 @@ import pl.llp.aircasting.database.migrations.*
         SensorThresholdDBObject::class,
         NoteDBObject::class,
         ActiveSessionMeasurementDBObject::class],
-    version = 29,
+    version = 30,
```
xameYeoh, line 21: Are you sure this solver the issue? I did migrate from previous version to the new one in my pr. It was a migration from 28 to 29

LinX64: Ohh I see. Need to check more...

xameYeoh: I will work on it, I know where the problem is now

LinX64: Ok it seems like it cannot find `stream_id` column.

LinX64:
Are you sure this solves the issue? Did you try to run those queries with the Database inspector?
Using the Database inspector, it seems like it happens again:

```
An error occurred while running the statement: index index_measurements_session_id_measurement_stream_id_time already exists (code 1 SQLITE_ERROR): , while compiling: CREATE UNIQUE INDEX `index_measurements_session_id_measurement_stream_id_time` ON `measurements` (`session_id`, `measurement_stream_id`, `time`)
```

xameYeoh:
It migrates correctly from 28 to 29, I've tried it manually on my device. If it crashes for you, when does it happen? If you try to run these queries on version 29, it will tell you that index already exists, because it was created already previously

LinX64:
> It migrates correctly from 28 to 29, I've tried it manually on my device. If it crashes for you, when does it happen? If this happens when try to run these queries manually on version 29, it will tell you that index already exists, because it was created already during migration (and the index is unique, it cannot be created twice)

Ok cool, let's test it then.

PR #431, opened 2022-04-22
Endpoints | APIService

LinX64:
Improvement

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/networking/Constants.kt
+++ b/app/src/main/java/pl/llp/aircasting/networking/Constants.kt
@@ -1,13 +1,13 @@
 package pl.llp.aircasting.networking
 
 object Constants {
-    /* GET Methods */
+    /* GET Requests */
     const val urlDownloadSession = "/api/user/sessions/empty.json"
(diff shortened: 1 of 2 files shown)


LinX64:
[Endpoints | APIService](https://trello.com/c/cMbgq0rL/1487-endpoints-apiservice)

xameYeoh (commented):
I guess that the idea was to have all API calls in one place, right? If this is true, they all already are in ApiService, are they not? Or are they used elsewhere? If that is the case, that's a cool idea. But if they are only used in the ApiService, I feel like having @GET(Constants.urlDownloadSession) instead of @GET("/api/user/sessions/empty.json") just adds another clicking layer or am I missing something?

LinX64:
> I guess that the idea was to have all API calls in one place, right? If this is true, they all already are in ApiService, are they not? Or are they used elsewhere? If that is the case, that's a cool idea. But if they are only used in the ApiService, I feel like having @get(Constants.urlDownloadSession) instead of @get("/api/user/sessions/empty.json") just adds another clicking layer or am I missing something?

Good question! It can hold all endpoints and URLs in a separated object class so that we can refer to it anytime we want. It's easier to add new features and develop applications while everything's separated and is in the right place.

PR #433, opened 2022-04-25
Bluetooth manager crash android 12

xameYeoh:
The pr contains separation for Bluetooth management for Devices with >= Android 12 and lower. They now use different BluetoothManager implementations which are separated from client code by an interface


app/src/main/java/pl/llp/aircasting/permissions/PermissionsManager.kt
```diff
@@ -20,7 +21,16 @@ open class PermissionsManager {
     private val AUDIO_PERMISSIONS = arrayOf(Manifest.permission.RECORD_AUDIO)
 
     private val BLUETOOTH_PERMISSIONS =
-        arrayOf(Manifest.permission.BLUETOOTH_CONNECT, Manifest.permission.BLUETOOTH_SCAN)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
LinX64, line 24: For the API versions check, you can use the extension functions available in `APIHELPER.kt`.

app/src/main/java/pl/llp/aircasting/permissions/PermissionsManager.kt
```diff
@@ -20,7 +21,17 @@ open class PermissionsManager {
...
+            arrayOf(
+                Manifest.permission.BLUETOOTH_CONNECT,
+                Manifest.permission.BLUETOOTH_SCAN
+            )
+        } else {
```
LinX64, line 29: Not important but, both `arrayOf`s are considered to be just one line so you can remove the curly brackets.

app/src/main/java/pl/llp/aircasting/di/PermissionsModule.kt
```diff
@@ -14,7 +18,10 @@ open class PermissionsModule {
 
     @Provides
     @Singleton
-    open fun providesBluetoothManager(permissionsManager: PermissionsManager): BluetoothManager {
-        return BluetoothManager()
+    open fun providesBluetoothManager(application: AircastingApplication): BluetoothManager {
+        return if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.S)
```
LinX64, line 22: Same for here. You could use the extension function.

LinX64 (approved):
Except for not using the extension function, all seem to be good. We can go for a test since there is no Android 12 device available.

PR #444, opened 2022-05-26
OnMarkers click handling + fix for the duplicate borders show

LinX64:
This is the implementation + fix for when user clicks on the marker and it selects the exact session and it removes the previous border from the cardView.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
@@ -1,29 +1,43 @@
 package pl.llp.aircasting.ui.view.adapters
 
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
+import androidx.core.content.ContextCompat
 import androidx.recyclerview.widget.RecyclerView
+import pl.llp.aircasting.R
 import pl.llp.aircasting.data.api.response.search.Session
 import pl.llp.aircasting.databinding.ItemSesssionsListFixedFollowBinding
 
 class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Unit) :
     RecyclerView.Adapter<FixedFollowAdapter.DataViewHolder>() {
     private val sessions: ArrayList<Session> = ArrayList()
+    private var selectedSession: Session? = null
+    private var sessionView: View? = null
 
     inner class DataViewHolder(
         private val binding: ItemSesssionsListFixedFollowBinding,
         private val onItemClicked: (Int) -> Unit
     ) : RecyclerView.ViewHolder(binding.root) {
-
         fun bind(sessions: Session) {
             binding.mySessions = sessions
             binding.executePendingBindings()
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -1,29 +1,43 @@
...
     private val sessions: ArrayList<Session> = ArrayList()
+    private var selectedSession: Session? = null
+    private var sessionView: View? = null
 
     inner class DataViewHolder(
         private val binding: ItemSesssionsListFixedFollowBinding,
         private val onItemClicked: (Int) -> Unit
     ) : RecyclerView.ViewHolder(binding.root) {
-
         fun bind(sessions: Session) {
```
xameYeoh, line 22: Should the parameter be "session" instead of "sessions", as it is only one?

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -1,29 +1,43 @@
...
             binding.mySessions = sessions
             binding.executePendingBindings()
 
             binding.root.setOnClickListener { onItemClicked(bindingAdapterPosition) }
+
+            sessionView = binding.root
+            binding.root.apply {
+                if (selectedSession?.id == sessions.id) setBackgroundResource(R.drawable.card_view_border_search) else setBackgroundColor(
```
xameYeoh, line 30: I think it would be nice to have two little private methods like:
val itemView = binding.root
if (selectedSession?.id == session.id) {
     setBackgroundWithBorder(itemView)
}
else setBackgroundWithoutBorder(itemView)

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -47,4 +61,21 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
         val session = sessions.first { it.uuid == uid }
         return sessions.indexOf(session)
     }
+
+    fun addCardBorder(position: Int) {
+        selectedSession = sessions[position]
+        removeItemBorder()
```
xameYeoh, line 67: I think we could provide more clarity with name like removeBorderFromPreviousCard()

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -47,4 +61,21 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
...
+        notifyItemChanged(position)
+    }
+
+    fun removeItemBorder() {
```
xameYeoh, line 71: I think we could provide more clarity with name like removeBorderFromPreviousCard()

xameYeoh (approved):

PR #446, opened 2022-05-26
Redo Button Implementation | Search & Follow

LinX64:


app/src/main/java/pl/llp/aircasting/data/local/entity/ExtSessionDBObject.kt
```diff
@@ -0,0 +1,22 @@
...
+    @ColumnInfo(name = "username") val username: String,
+    @ColumnInfo(name = "endTimeLocal") val endTimeLocal: String,
+    @ColumnInfo(name = "startTimeLocal") val startTimeLocal: String,
+    @ColumnInfo(name = "lastHourAverage") val lastHourAverage: Double,
+    @ColumnInfo(name = "latitude") val latitude: Double,
+    @ColumnInfo(name = "longitude") val longitude: String,
+    @ColumnInfo(name = "isIndoor") val isIndoor: Boolean,
+    //@ColumnInfo(name = "streams") val streams: Streams,
+)
+// TODO: Need to do sth with the streams since we cannot save it like that on DB
```
xameYeoh, line 22: I guess we will need to create another table for the streams /
Or if the streams are similar to the ones that we have in DB already - we could convert them;
And in either case here would be an ID of the stream that will reference it in the other Streams table

xameYeoh (commented):
Left a comment

LinX64:
Mergin this PR to be able to work on the DB part in just one PR.

PR #453, opened 2022-06-01
Chip Groups Selection Fix

LinX64:
This fixes the issue with the chip groups which was mistakenly being handled in one callback method.


app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -54,8 +53,27 @@ class SearchFixedSessionsActivity : AppCompatActivity(), ChipGroup.OnCheckedStat
         setSupportActionBar(topAppBar)
         supportActionBar?.setDisplayHomeAsUpEnabled(true)
 
-        binding.chipGroupFirst.setOnCheckedStateChangeListener(this)
-        binding.chipGroupSensors.setOnCheckedStateChangeListener(this)
+        binding.chipGroupFirst.setOnCheckedStateChangeListener { chipGroup, checkedIds ->
```
xameYeoh, line 56: Is there a way to rename it like binding.chipGroupParent?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -54,8 +53,27 @@ class SearchFixedSessionsActivity : AppCompatActivity(), ChipGroup.OnCheckedStat
...
+                binding.airbeamChip.gone()
+                binding.purpleAirChip.gone()
+                Ozone.OPEN_AQ.getMeasurementType()
+            } else {
+                binding.airbeamChip.visible()
+                binding.purpleAirChip.visible()
+                ParticulateMatter.AIRBEAM.getMeasurementType()
+            }
+        }
+        binding.chipGroupSensors.setOnCheckedStateChangeListener { chipGroup, checkedIds ->
```
xameYeoh, line 67: And this one to binding.chipGroupSecondLevel e.g.

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -54,8 +53,27 @@ class SearchFixedSessionsActivity : AppCompatActivity(), ChipGroup.OnCheckedStat
...
+        binding.chipGroupFirst.setOnCheckedStateChangeListener { chipGroup, checkedIds ->
+            txtSelectedParameter = if (chipGroup.checkedChipId == binding.ozoneChip.id) {
+                binding.airbeamChip.gone()
+                binding.purpleAirChip.gone()
+                Ozone.OPEN_AQ.getMeasurementType()
+            } else {
+                binding.airbeamChip.visible()
+                binding.purpleAirChip.visible()
+                ParticulateMatter.AIRBEAM.getMeasurementType()
+            }
```
xameYeoh, line 65: I also think that we may have a FirstLevelChip group (for ParticulateMatter and Ozone) and then have a two separate ChipGroups for ParticulateMatter sensors and Ozone sensors (so not 2 groups as we have now, but three) - and the FirstLevel ChipGroup will only be responsible for hiding/unhiding 2nd and 3rd ChipGroups. While 2nd level ChipGroups will be responsible for setting the txtSelectedParameter. I think this structure would be better, because if in theory there will be a third FirstLevel ChipGroup, we will have juggle all those individual chips / or for e.g. if the sensor name will change and Ozone sensor will not be Open AQ, but something else - and we will not be able to utilise the same Open AQ chip from ParticulateMatter and Ozone

xameYeoh: In this way, it will be easier and more cleaner to set up click handlers in code as well, as FirstLevel will only hide/unhide SecondLevel ChipGroups, while SecondLevel chips will all correspond to one SensorType from the enum

xameYeoh: Let me know if this is unclear, 'cause it's a bit hard to describe :D

xameYeoh (commented):
I think that the ChipsGroups structure could be a bit different to be more reusable

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -37,7 +37,7 @@ class SearchFixedSessionsActivity : AppCompatActivity() {
 
     private lateinit var binding: ActivitySearchFixedSessionsBinding
     private var placesClient: PlacesClient? = null
-    private var txtSelectedParameter: String? = null
+    private var txtSelectedParameter: String = ParticulateMatter.AIRBEAM.getMeasurementType()
```
xameYeoh, line 40: Good catch setting up the default value! 👍

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -53,27 +53,32 @@ class SearchFixedSessionsActivity : AppCompatActivity() {
...
-                binding.airbeamChip.visible()
-                binding.purpleAirChip.visible()
-                ParticulateMatter.AIRBEAM.getMeasurementType()
+        binding.chipGroupFirstLevel.setOnCheckedStateChangeListener { chipGroup, _ ->
+            if (chipGroup.checkedChipId == binding.ozoneChip.id) binding.apply {
+                chipGroupSecondLevel.gone()
+                chipGroupThirdLevel.visible()
+            } else binding.apply {
+                chipGroupSecondLevel.visible()
+                chipGroupThirdLevel.gone()
```
xameYeoh, line 62: I would have the naming like chipGroupSecondLevelOne and chipGroupSecondLevelTwo (as they are both on the same chip level), what do you think?

xameYeoh (commented):
This looks awesome! 🎉

xameYeoh (commented):
This looks awesome! 🎉

PR #465, opened 2022-06-07
Search location | Search and follow

LinX64:
This addes the second search functionality to the project.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
@@ -54,12 +54,8 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
         holder.bind(sessions[position])
 
     fun refresh(list: List<Session>) {
-        clearAdapter()
-        sessions.addAll(list)
-    }
-
-    fun clearAdapter() {
         sessions.clear()
+        sessions.addAll(list)
     }
 
     fun getSessionPositionBasedOnId(uid: String): Int {
--- a/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
@@ -27,7 +27,9 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
                 setOnClickListener { onItemClicked(bindingAdapterPosition) }
                 cardView = this
 
-                if (selectedSession?.id == session.id) setBackgroundWithBorder(cardView) else setBackgroundWithoutBorder(cardView)
+                if (selectedSession?.id == session.id) setBackgroundWithBorder(cardView) else setBackgroundWithoutBorder(
+                    cardView
+                )
             }
         }
     }
@@ -54,6 +56,10 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
     fun refresh(list: List<Session>) {
         sessions.clear()
         sessions.addAll(list)
+    }
+
+    fun clearAdapter() {
+        sessions.clear()
     }
 
     fun getSessionPositionBasedOnId(uid: String): Int {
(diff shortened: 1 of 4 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -56,6 +58,10 @@ class FixedFollowAdapter constructor(private val onItemClicked: (Session) -> Uni
         sessions.addAll(list)
     }
 
+    fun clearAdapter() {
+        sessions.clear()
+    }
+
```
xameYeoh, line 64: I think the clearing in the adapter happens every time we add new sessions to it:
```
    fun refresh(list: List<Session>) {
        sessions.clear()
        sessions.addAll(list)
    }
```

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -216,6 +263,16 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
         )
     }
 
+    private fun secondSearchSetup(lat: Double, long: Double) {
+        mMap.clear()
+
+        val theLocation = LatLng(lat, long)
+        mMap.addMarker(options.position(theLocation))
```
xameYeoh, line 270: I thought Michael does not want the red marker on the searched location, or is it another marker?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -105,6 +113,47 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
...
+    }
+
+    private fun initialisePlacesClient() {
+        if (!Places.isInitialized()) Places.initialize(
+            applicationContext,
+            BuildConfig.PLACES_API_KEY
+        )
+        placesClient = Places.createClient(this)
+    }
+
```
xameYeoh, line 156: Good extraction! 👍

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -216,6 +263,16 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
         )
     }
 
+    private fun secondSearchSetup(lat: Double, long: Double) {
```
xameYeoh, line 266: I would name it like moveMapToSelectedLocationAndRefresh() . I think it is always helpful to explicitly state what a method is doing, then it requires less time to navigate through

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -216,6 +263,16 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
...
+        mMap.clear()
+
+        val theLocation = LatLng(lat, long)
```
xameYeoh, line 269: I think selectedLocation name would be better

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
```diff
@@ -138,6 +134,14 @@ class SearchFixedSessionsActivity : AppCompatActivity() {
         }
     }
 
+    private fun initialisePlacesClient() {
+        if (!Places.isInitialized()) Places.initialize(
+            applicationContext,
+            BuildConfig.PLACES_API_KEY
+        )
+        placesClient = Places.createClient(this)
+    }
```
xameYeoh, line 143: If we need to initialise it at different parts of code, is there a possibility to extract it somewhere to not define the private method several times? So that everywhere we would have: 
placesClient = initialise(placesClient)

xameYeoh (changes requested):
I've added several comments. We can cooperate on the refactoring if something is unclear, or you have better ideas :)

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionsActivity.kt
LinX64: I thought the same, but turns out it needs to be initialized for each activity separately after `onCreate()`, so instead, I've added it as an extension function.

xameYeoh (approved):

PR #466, opened 2022-06-08
Maps: UI improvements

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -206,6 +253,16 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
...
+    private fun getSensor(): String {
+        return when (txtSensor) {
+            "airbeam2-pm2.5" -> "AirBeam"
+            "openaq-pm2.5" -> "OpenAQ"
+            "purpleair-pm2.5" -> "PurpleAir"
+            "openaq-o3" -> "Ozone"
+            else -> "AirBeam"
+        }
+    }
+
```
xameYeoh, line 265: Can't we somehow tie this to the Enum that we have so that in case we need to change the strings we will only need to change them in Enum?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
```diff
@@ -129,14 +128,20 @@ class SearchFixedBottomSheet : BottomSheet(), OnMapReadyCallback {
...
 
+    private fun drawMarkerOnMap(lat: Double, lng: Double): Marker? {
+        return mMap.addMarker(
+            MarkerOptions()
+                .position(LatLng(lat, lng))
+                .anchor(0.5f, 0.5f)
+                .icon(bitmapDescriptorFromVector(requireActivity(), R.drawable.ic_dot_20))
+        )
+    }
+
```
xameYeoh, line 144: Maybe we can somehow unify this method and extract it from the class, as it is being used throughout the codebase so that we can have:
drawMarkerOnMap(map, lat, lng, markerInfo?) 
And have Marker info for the uuid?

xameYeoh (changes requested):
I've left a few comments on possible directions of refactoring

app/src/main/java/pl/llp/aircasting/data/api/util/SensorName.kt
```diff
@@ -0,0 +1,5 @@
+package pl.llp.aircasting.data.api.util
+
+interface SensorName {
+    fun getSensorName(): String
```
xameYeoh, line 4: I think we should have another name for these methods, not to confuse them with the first Enum. Something like getNameToDisplay()

LinX64:
[S&F > maps: UI improvements](https://trello.com/c/AyLVgZkt/1539-sf-maps-ui-improvements)

PR #468, opened 2022-06-09
Saving external sessions

xameYeoh:
This is an approach to save the sessions and measurement streams without adding new tables to the DB, but instead converting the api response sessions to the format we have in DB and using new sessions and streams as old ones


app/src/main/java/pl/llp/aircasting/data/local/repository/SessionsRepository.kt
```diff
@@ -16,7 +16,7 @@ class SessionsRepository {
         return mDatabase.sessions().insert(sessionDBObject)
     }
 
-    fun insert(session: pl.llp.aircasting.data.api.response.search.Session): Long {
+    fun insert(session: pl.llp.aircasting.data.api.response.search.SessionInRegionResponse): Long {
```
LinX64: Shouldn't we import this? 

`import pl.llp.aircasting.data.api.response.search.SessionInRegionResponse`

app/src/main/java/pl/llp/aircasting/data/model/Session.kt
```diff
@@ -78,7 +78,7 @@ open class Session(
         this.locationless = locationless
     }
 
-    constructor(apiSession: pl.llp.aircasting.data.api.response.search.Session) : this(
+    constructor(apiSession: pl.llp.aircasting.data.api.response.search.SessionInRegionResponse) : this(
```
LinX64, line 81: Same here -> `import pl.llp.aircasting.data.api.response.search.SessionInRegionResponse`

LinX64 (approved):

PR #471, opened 2022-06-14
Crash affecting one user with Android 12 devices

xameYeoh:
Made title nullable (as we have sessions with nil titles in the backend); Added "" fallback value during the initialization


LinX64 (approved):
This fixes the crash but for showing to the user, it will show one session without a name that doesn't seem to be good for user experience. The other thing is that the `mName` parameter is a non-nullable object. I'm not sure how this happened and the session was created without a name. We should probably add a name to it on the backend side or on ours (Android/IOS)

PR #473, opened 2022-06-15
Chart on the BottomSheet

LinX64:


LinX64:
[5. [Session] Chart](https://trello.com/c/3lmXkMuQ/1536-5-session-chart)

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -25,13 +25,13 @@ class SessionPresenter() {
...
         loading: Boolean = false
-    ): this() {
+    ) : this() {
         this.session = session
         this.selectedStream = selectedStream ?: defaultStream(session)
         this.expanded = expanded
         this.loading = loading
         this.sensorThresholds = sensorThresholds
-        if (session.tab == SessionsTab.FOLLOWING || session.tab == SessionsTab.MOBILE_ACTIVE) {
+        if (session.tab == SessionsTab.FOLLOWING || session.tab == SessionsTab.MOBILE_ACTIVE && session.isExternal) {
```
xameYeoh, line 34: Should it not be OR session.isExternal ?

app/src/main/java/pl/llp/aircasting/data/model/SensorThreshold.kt
```diff
@@ -19,10 +20,26 @@ class SensorThreshold(
...
 
-    val levels get() = arrayOf(thresholdVeryLow, thresholdLow, thresholdMedium, thresholdHigh, thresholdVeryHigh)
+    val levels
+        get() = arrayOf(
+            thresholdVeryLow,
+            thresholdLow,
+            thresholdMedium,
+            thresholdHigh,
+            thresholdVeryHigh
+        )
```
xameYeoh, line 42: I did not see this getter used in this pr, but I presume that's for the SessionPresenter, right? I think this is a good idea to be able to get all the thresholds from the session in some sort of single structure. But I would use the SensorThreshold as a return value from here - as this is precisely what SessionPresenter needs to be constructed. So we might as well have a getter for SensorThreshold getter from these thresholds, instead of a list

xameYeoh (changes requested):
I've highlighted the parts that I think require some refactoring

app/src/main/java/pl/llp/aircasting/data/model/SensorThreshold.kt
LinX64: Good point! I don't think it is for the `SessionPresenter` as I saw it was being used for the `Measurement` and for getting the levels and showing it with the specific colors. As we discussed, the `SessionPresenter` already has those parameters in its own constructor so I guess that we can use the same approach I used without any problems.

LinX64:
[Improvements](https://trello.com/c/etqvBVnj/1565-improvements)

app/src/main/java/pl/llp/aircasting/data/model/SensorThreshold.kt
xameYeoh: Sorry, I confused it a bit - I thought this was in a MeasurementStream class. I guess it would be nice for this class to provide it's thresholds in a bundle as SensorThreshold, but if all other parts of the app construct it the same way as you do for the SessionPresenter, I guess we better have it unified

PR #475, opened 2022-06-16
Save thresholds external sessions

xameYeoh:
The pr saves all the external sessions' information to the database on follow button click. It also uses the new endpoint

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
@@ -9,7 +9,6 @@ import pl.llp.aircasting.data.api.response.search.session.details.SessionWithStr
 import pl.llp.aircasting.data.api.util.SensorInformation
 import pl.llp.aircasting.data.local.repository.*
 import pl.llp.aircasting.data.model.*
-import pl.llp.aircasting.di.modules.DefaultDispatcher
 import pl.llp.aircasting.di.modules.IoDispatcher
 import pl.llp.aircasting.util.Resource
 import javax.inject.Inject
@@ -21,8 +20,7 @@ class SearchFollowViewModel @Inject constructor(
     private val measurementStreamsRepository: MeasurementStreamsRepository,
     private val sessionsRepository: SessionsRepository,
     private val thresholdsRepository: ThresholdsRepository,
-    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
-    @DefaultDispatcher private val defaultDispatcher: CoroutineDispatcher
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher
 ) : ViewModel() {
     private val mutableSelectedSession = MutableLiveData<SessionInRegionResponse>()
     private val mutableLat = MutableLiveData<Double>()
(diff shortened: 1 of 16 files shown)


app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
```diff
@@ -43,8 +38,36 @@ class SearchFollowViewModel @Inject constructor(
 
     fun selectSession(session: SessionInRegionResponse) {
         mutableSelectedSession.value = session
+
+        selectedSessionWithStreamsResponse = downloadFullSessionAsync(session)
+        selectedFullSession = initializeModelFromResponseAsync()
```
LinX64, line 43: You can use [`awaitAll()`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/Await.kt#L41) to wait for both calls. A great example:  https://developer.android.com/kotlin/coroutines/coroutines-adv#parallel

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
```diff
@@ -43,8 +38,36 @@ class SearchFollowViewModel @Inject constructor(
...
 
+    private fun downloadFullSessionAsync(session: SessionInRegionResponse) =
+        viewModelScope.async(ioDispatcher) {
+            activeFixedRepo.getSessionWithStreamsAndMeasurements(
+                session.id
+            )
+        }
+
+    private fun initializeModelFromResponseAsync(): Deferred<Session?> =
+        viewModelScope.async(defaultDispatcher) {
```
LinX64, line 54: You don't really need to specify an already added dispatcher here:

> Coroutine context is inherited from a CoroutineScope, additional context elements can be specified with context argument. **If the context does not have any dispatcher nor any other ContinuationInterceptor, then `Dispatchers.Default` is used**. The parent job is inherited from a CoroutineScope as well, but it can also be overridden with corresponding context element.

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
```diff
@@ -57,67 +80,86 @@ class SearchFollowViewModel @Inject constructor(
...
+        streamId: Long,
+        sessionId: Long,
+        stream: MeasurementStream
+    ) {
+        measurementsRepository.insertAll(streamId, sessionId, stream.measurements)
+        activeSessionMeasurementsRepository.insertAll(streamId, sessionId, stream.measurements)
+    }
+
+    fun deleteSession(
         session: SessionInRegionResponse,
```
LinX64, line 163: Are you gonna add another parameter to this method? Otherwise having another "," doesn't make any sense.

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
```diff
@@ -151,30 +193,4 @@ class SearchFollowViewModel @Inject constructor(
             )
             emit(stream)
         }
-
-    private suspend fun getMeasurementsFromSelectedSession(): List<Measurement> {
-        val sessionId = selectedSession.value?.id?.toLong()
```
LinX64, line 156: `toLong()` can be removed since you changed the id type to Long.

LinX64 (changes requested):
I've added a few comments regarding the changes which should be considered.

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
xameYeoh: I see you have a grudge against trailing commas 😄

LinX64 (approved):
Looks good.

PR #480, opened 2022-06-21
Chart external sessions

xameYeoh:
Making chart work correctly regardless of rate of measurements in a session


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/ChartAveragesCreator.kt
```diff
@@ -10,10 +10,9 @@ import kotlin.math.roundToInt
 class ChartAveragesCreator {
     companion object {
         const val MAX_AVERAGES_AMOUNT = 9
-        const val NUMBER_OF_MEASUREMENTS_IN_ONE_AVERAGE = 60
         private val MOBILE_INTERVAL_IN_SECONDS = 60
         private const val MAX_X_VALUE = 8.0
-        private const val MIN_X_VALUE = 0.0
+        private const val MIN_X_VALUE = 0
         private val MOBILE_FREQUENCY_DIVISOR = 8 * 1000.toDouble()
```
LinX64, line 16: You can make this as constant.

LinX64 (approved):
If it works, then it works. Nice job! Separating that ext - internal sessions was a great idea!

PR #481, opened 2022-06-22
Map wrong location | Improvements

LinX64:


LinX64:
[Map wrong location on BottomSheet | Improvements](https://trello.com/c/FPWIutEZ/1566-map-wrong-location-on-bottomsheet-improvements)

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
```diff
@@ -85,11 +89,13 @@ class SearchFixedBottomSheet : BottomSheet(), OnMapReadyCallback {
...
-            if (searchFollowViewModel.isSelectedSessionFollowed.await())
-                toggleUnFollowButton()
-            else
-                toggleFollowButton()
+        searchFollowViewModel.apply {
+            viewModelScope.launch {
+                if (isSelectedSessionFollowed.await())
+                    toggleUnFollowButton()
+                else
+                    toggleFollowButton()
```
xameYeoh, line 97: Doesn't it require launch on MainThread, as we are updating the UI?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -233,30 +233,16 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
...
-    private fun passLatLng() {
-        val selectedLat = lat?.toDouble()
-        val selectedLng = lng?.toDouble()
-        if (selectedLat != null && selectedLng != null) {
-            searchFollowViewModel.getLat(selectedLat)
-            searchFollowViewModel.getLng(selectedLng)
-        }
+    private fun passLatLng(session: SessionInRegionResponse) {
+        searchFollowViewModel.getLat(session.latitude)
+        searchFollowViewModel.getLng(session.longitude)
```
xameYeoh, line 245: I see there are methods like that in the ViewModel: 
`fun getLat(lat: Double) {
        mutableLat.value = lat
    }`
    Should not they be named setLat and setLng then, as they are setting the values?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
LinX64: The `default` launch will run on the mainThread, so you basically don't need to specify it as it is already being launched on the mainThread and we're launching it from the `view`.

xameYeoh: I thought that it uses Dispatchers.Default by default, not Dispatchers.Main. But I've just read that the viewModelScope uses Dispatchers.Main as a default one. Learning something new 😄👍

PR #482, opened 2022-06-22
AirBeam3 Sessions and Measurements

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
```diff
@@ -109,11 +118,11 @@ class SearchFixedBottomSheet : BottomSheet(), OnMapReadyCallback {
         binding?.unfollowBtn?.visible()
     }
 
-    private fun setupLoader() {
+    /*private fun setupLoader() {
         val loaderImage =
             binding?.measurementsTableBinding?.streamMeasurementHeaderAndValue?.loaderImage as ImageView
         loader = AnimatedLoader(loaderImage)
-    }
+    }*/
```
xameYeoh, line 125: If the loader is present somewhere in the logic on measurementsTableContainer, I guess there is no point in keeping the commented out part and it can be deleted

PR #483, opened 2022-06-24
Disable follow feature for your sessions

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
```diff
@@ -105,8 +106,20 @@ class SearchFixedBottomSheet : BottomSheet(), OnMapReadyCallback {
...
 
     private fun toggleUnFollowButton() {
-        binding?.followBtn?.gone()
-        binding?.unfollowBtn?.visible()
+        if (!searchFollowViewModel.checkIfIsOwnSession()) {
+            binding?.followBtn?.gone()
+            binding?.unfollowBtn?.visible()
+        } else toggleOwnSessionFollowButton()
+    }
+
```
xameYeoh, line 114: There is a method toggleCorrectButton(); can these checks be transferred there somehow? So that for toggleButton methods we will just have the UI logic

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
```diff
@@ -34,16 +36,23 @@ class SearchFollowViewModel @Inject constructor(
     val myLng: LiveData<Double> get() = mutableLng
     val thresholdColor: LiveData<Int> get() = mutableThresholdColor
     lateinit var isSelectedSessionFollowed: Deferred<Boolean>
+    private var isOwnSession: Boolean = false
 
     fun selectSession(session: SessionInRegionResponse) {
         mutableSelectedSession.value = session
 
+        if (session.username == mSettings.getProfileName()) isOwnSession = true
+
```
xameYeoh, line 45: I think the action could be transferred to checkIfUserOwnsSession() method, as this is what the name of the method implies; so that we will have isOwnSession = checkIfUserOwnsSession() here

xameYeoh: And for getting the value in bottomsheet, we could have a get() getter

xameYeoh (commented):
I've left a few comments to consider

app/src/main/java/pl/llp/aircasting/ui/viewmodel/SearchFollowViewModel.kt
LinX64: I also thought about it at first, but you should consider the `session` which needs to be passed to that `checkIfUserOwnsSession()` method. I actually ended up doing it in another way, so we now have the `checkIfUserOwnsSession()` method, which returns the condition value, then returning the `isOwnSession` value from a separate method `userOwnsSession()` which can be used from the view.

PR #484, opened 2022-06-28
Chart Update | BottomSheet

LinX64:


LinX64:
[The chart is not updating on the BottomSheet while being selected by User](https://trello.com/c/jcEEYw2N/1571-the-chart-is-not-updating-on-the-bottomsheet-while-being-selected-by-user)

PR #486, opened 2022-06-29
ProgressBar on the BottomSheet

LinX64:


LinX64:
[ProgressBar for when the measurements are being loaded on the BottomSheet](https://trello.com/c/WrEEyvKH/1575-progressbar-for-when-the-measurements-are-being-loaded-on-the-bottomsheet)

app/src/main/java/pl/llp/aircasting/AircastingApplication.kt
```diff
@@ -41,6 +39,11 @@ class AircastingApplication : Application() {
         ProcessLifecycleOwner.get()
             .lifecycle
             .addObserver(AppLifecycleObserver())
+    }
 
+    private fun checkIfThemeChangeEnabled() {
+        if (mSettings?.isThemeChangeEnabled() == true) AppCompatDelegate.setDefaultNightMode(
+            AppCompatDelegate.MODE_NIGHT_YES
+        ) else AppCompatDelegate.setDefaultNightMode(AppCompatDelegate.MODE_NIGHT_NO)
```
xameYeoh, line 47: As this method sets the theme (not just checks it), I would call it setCorrectAppTheme()

xameYeoh (changes requested):
One naming

PR #489, opened 2022-06-29
External sessions chart times

xameYeoh:
Refactor of timestamps for chart with regard to UTC time of external sessions

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/ExternalSessionChartDataCalculator.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/ExternalSessionChartDataCalculator.kt
@@ -4,21 +4,23 @@ import com.github.mikephil.charting.data.Entry
 import pl.llp.aircasting.data.api.util.StringConstants
 import pl.llp.aircasting.data.model.MeasurementStream
 import pl.llp.aircasting.data.model.Session
+import java.util.*
 
 class ExternalSessionChartDataCalculator(session: Session) : SessionChartDataCalculator(session) {
-
     override fun calculateEntriesAndTimestamps(stream: MeasurementStream?): MutableList<Entry>? {
+        val timeStampsSetter = UTCTimeStampsSetter()
+
         return when {
             stream == null -> null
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/ExternalSessionChartDataCalculator.kt
```diff
@@ -4,26 +4,34 @@ import com.github.mikephil.charting.data.Entry
...
 import pl.llp.aircasting.data.model.MeasurementStream
 import pl.llp.aircasting.data.model.Session
+import java.util.*
 
 class ExternalSessionChartDataCalculator(session: Session) : SessionChartDataCalculator(session) {
-
     override fun calculateEntriesAndTimestamps(stream: MeasurementStream?): MutableList<Entry>? {
+        val timeStampsSetter = UTCTimeStampsSetter()
+
         return when {
```
LinX64, line 13: Couldn't we have a cleaner and more understanding expression here? It's returning `getFixedEntries()` in the `else` block. Can't we return the `null` part(`stream == null -> null`) in the `else` block?

xameYeoh: That is, if you are speaking about this part:
`override fun calculateEntriesAndTimestamps(stream: MeasurementStream?): MutableList<Entry>? {
        val timeStampsSetter = UTCTimeStampsSetter()

        return when {
            stream == null -> null

            isFromOpenAQ(stream) -> OpenAQChartAveragesCreator().getFixedEntries(
                stream,
                timeStampsSetter
            )

            else -> ChartAveragesCreator().getFixedEntries(
                stream,
                timeStampsSetter
            )
        }
    }`

xameYeoh: Check the last commit if it looks better :)

LinX64: You got it!

PR #491, opened 2022-06-29
30 31 migration fix

xameYeoh:
Changing the migration from previous version of DB, as we changed our modification on the DB

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/data/local/entity/sessions.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/local/entity/sessions.kt
@@ -34,7 +34,7 @@ data class SessionDBObject(
     @ColumnInfo(name = "averaging_frequency") val averaging_frequency: Int = 1,
     @ColumnInfo(name = "session_order") val session_order: Int? = null,
     @ColumnInfo(name = "username") val username: String? = null,
-    @ColumnInfo(name = "is_external") val isExternal: Boolean = false
+    @ColumnInfo(name = "is_external", defaultValue = "0") val isExternal: Boolean? = false
 ) {
     @PrimaryKey(autoGenerate = true)
     var id: Long = 0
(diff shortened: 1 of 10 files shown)


app/src/main/java/pl/llp/aircasting/data/local/entity/sessions.kt
```diff
@@ -34,7 +34,7 @@ data class SessionDBObject(
     @ColumnInfo(name = "averaging_frequency") val averaging_frequency: Int = 1,
     @ColumnInfo(name = "session_order") val session_order: Int? = null,
     @ColumnInfo(name = "username") val username: String? = null,
-    @ColumnInfo(name = "is_external") val isExternal: Boolean = false
+    @ColumnInfo(name = "is_external", defaultValue = "0") val isExternal: Boolean? = false
```
LinX64, line 37: I understand that `"0"` means `false`, but the thing is that you are already setting `false` on the Kotlin side, and based on [the information](https://i.imgur.com/6QtEnzF.png), `defaultValue` will not work if we use `@Insert`, which we are: 

```
 @Insert(onConflict = OnConflictStrategy.IGNORE)
    fun insert(session: SessionDBObject): Long
```

Setting `defaultValue` in this case with adding `false` on the Kotlin side seems to be useless or maybe I am missing something?

xameYeoh: As far as I got it, the default value is needed because we have to set something for rows which already exist on users side, otherwise migration does not pass. The thing is that when a table already exists and you add a new column there, you cannot have this column as NonNull in SQLite. So that's why we have to make it nullable, but then we have to provide default value for existing rows, which will receive it during the migration. This is only related to the SQL side. The default value "false" in Kotlin will be used from the app's side, when you instantiate the object for insertion, and don't provide a specific value for isExternal. This two default values are for different purposes, I believe

PR #506, opened 2022-07-04
Cut measurements to 24h after download

xameYeoh:


LinX64 (approved):

app/src/main/java/pl/llp/aircasting/data/model/MeasurementStream.kt
```diff
@@ -227,6 +229,12 @@ open class MeasurementStream(
         return allMeasurements.subList(measurementsSize - amount, measurementsSize)
     }
 
+    fun getLast24HoursOfMeasurements(): List<Measurement> {
+        val end = measurements.last().time
+        val start = calendar().addHours(end, -24)
+        return measurements.filter { it.time in start..end }
```
LinX64, line 235: Looks good! You could also add clarifying parenthesis for the start and end; 

`filter { it.time in (start..end) }`

LinX64 (approved):

app/src/main/java/pl/llp/aircasting/data/model/MeasurementStream.kt
LinX64: I already know that it is included in Kotlin, that's why I suggested it, right? :)

Now with the variable option, it looks better.

PR #507, opened 2022-07-04
Improvements for the card item selection

LinX64:
This should fix the problems with card item selection and texts overlaps with the items on the card.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
@@ -310,7 +310,7 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
         val position = adapter.getSessionPositionBasedOnId(uuid)
 
         binding.recyclerFixedFollow.scrollToPosition(position)
-        adapter.addCardBorder(position)
+        adapter.scrollToSelectedCard(position)
     }
 
     override fun onMarkerClick(marker: Marker): Boolean {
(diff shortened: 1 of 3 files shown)


LinX64:
[S&F > map > cards: increase vertical space between session title & timestamp](https://trello.com/c/n7BrSYyL/1591-sf-map-cards-increase-vertical-space-between-session-title-timestamp)

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -25,20 +26,22 @@ class FixedFollowAdapter constructor(private val onItemClicked: (SessionInRegion
             binding.mySessions = session
             binding.executePendingBindings()
 
-            binding.root.apply {
+            binding.cvSessions.apply {
```
xameYeoh, line 29: I am not sure about whether it is a common abbreviation in Android world, if it is - let's have it like that. It's not that obvious to me, I would name it like: binding.sessionCard

LinX64: No, it's not. Now thanks to Databinding, we can bind the data in the correct and exact component.

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -25,20 +26,22 @@ class FixedFollowAdapter constructor(private val onItemClicked: (SessionInRegion
...
                 setOnClickListener {
                     onItemClicked(bindingAdapterPosition)
-
-                    // disable for a second to prevent double click
                     disableForASecond()
```
xameYeoh, line 32: I think it is better to leave the explanation, in case some developer in the future decides to remove this disableForASecond() and will have to spend time figuring out why it crashes. I would leave the following comment: "prevents duplicate fragment transaction for bottom sheet"

LinX64: Makes sense!

xameYeoh (approved):

PR #512, opened 2022-07-06
Settings: Satellite view option

LinX64:


app/src/main/java/pl/llp/aircasting/util/extensions.kt
```diff
@@ -221,3 +221,17 @@ fun Calendar.addHours(time: Date, hours: Int): Date {
...
+
+fun GoogleMap.setMapTypeToNormalWithStyle(mContext: Context) {
+    this.mapType = GoogleMap.MAP_TYPE_NORMAL
+    styleGoogleMap(this, mContext)
+}
+
+fun GoogleMap.checkIfSatelliteViewIsEnabled(mSettings: Settings, mContext: Context) {
+    if (mSettings.isUsingSatelliteView()) this.setMapTypeToSatellite()
+    else this.setMapTypeToNormalWithStyle(mContext)
+}
```
xameYeoh, line 237: I think this method sets the appropriate style of the map based on settings, so I would name it just "GoogleMap.setStyle()"

xameYeoh (changes requested):
I've left one comment to consider

PR #513, opened 2022-07-06
Saving expanded to prefs

xameYeoh:


LinX64 (changes requested):
Looks good, but I believe that you missed the logout part where it should clean all those values from the SharedPreference/StringSet when the user logs out as you mentioned in the relevant card.

Take a look at the `deleteFromSettings()` method in `Settings.kt` class.

LinX64 (approved):

PR #514, opened 2022-07-06
Map: Session marker doesn't shrink

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -94,4 +96,10 @@ class FixedFollowAdapter constructor(private val onItemClicked: (SessionInRegion
             strokeColor = ContextCompat.getColor(context, R.color.aircasting_blue_400)
         }
     }
+
+    private fun setBorderAndDotToSelectedSessionCard(position: Int, sessionID: Long) {
+        scrollToSelectedCard(position)
+
+        (cardView.context as SearchFixedResultActivity).highlightTheSelectedDot(sessionID.toString())
+    }
```
xameYeoh, line 104: Seems like this method scrolls to selected item and highlights marker. So I would name it just like that: scrollToSelectedItemAndHighlightItsMarker()

app/src/main/java/pl/llp/aircasting/ui/view/adapters/FixedFollowAdapter.kt
```diff
@@ -70,8 +72,8 @@ class FixedFollowAdapter constructor(private val onItemClicked: (SessionInRegion
         sessions.addAll(list)
     }
 
-    fun getSessionPositionBasedOnId(uid: String): Int {
-        val session = sessions.first { it.uuid == uid }
+    fun getSessionPositionBasedOnId(id: Long): Int {
+        val session = sessions.first { it.id == id }
```
xameYeoh, line 76: Maybe consider going back to UUID, as it is used everywhere to compare the session and is unique 100%

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -296,13 +299,23 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
     }
 
     private fun selectCorrespondingCardView(marker: Marker) {
-        val uuid = marker.snippet.toString()
-        val position = adapter.getSessionPositionBasedOnId(uuid)
+        val id = marker.snippet.toString()
+        val position = adapter.getSessionPositionBasedOnId(id.toLong())
```
xameYeoh, line 303: I would have id = marker.snippet.toString().toLong(), so that the value is ready to be used in getSessionPositionBasedOnId()

xameYeoh (changes requested):
I've left some comments to consider

PR #515, opened 2022-07-06
Session Card and Dot Issue

LinX64:


LinX64:
[Session Card and Dot Issue](https://trello.com/c/8VLW8upp/1603-session-card-and-dot-issue)

PR #519, opened 2022-07-07
Satellite View Should show details 

LinX64:


LinX64:
[maps: sattelite view should show the same place names as the default map view, e.g. street names](https://trello.com/c/HU3wiJrk/1608-maps-sattelite-view-should-show-the-same-place-names-as-the-default-map-view-eg-street-names)

xameYeoh (approved):

PR #521, opened 2022-07-07
Glued ab2 and ab3 responses in repository

xameYeoh:


LinX64 (approved):
Looks good, we'll have to change the implementation later.

PR #523, opened 2022-07-07
Set Pm2.5 as default

LinX64:


LinX64:
[Make Pm2.5 default for both external sessions and internal](https://trello.com/c/Q2W3DDsQ/1610-make-pm25-default-for-both-external-sessions-and-internal)

app/src/main/java/pl/llp/aircasting/data/api/util/StringConstants.kt
```diff
@@ -27,6 +27,8 @@ object StringConstants {
     const val purpleAir = "PurpleAir"
     const val openAQ = "OpenAQ"
 
+    const val pmTwoPointFive = "PM2.5"
+
```
xameYeoh, line 31: As far as I got it, the streamsSortedByDetailedType() works with AirBeamSensorName enum, which has detailedType strings in it, including "PM2.5". I would use the one from there, instead of creating another one, so that it will be from one source

xameYeoh: And in this case we can rename the enum from AirBeamSensorName to just SensorName, as we work not only with airbeams now

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -105,7 +106,7 @@ class SessionPresenter() {
 
     companion object {
         fun defaultStream(session: Session?): MeasurementStream? {
-            return session?.streamsSortedByDetailedType()?.firstOrNull()
+            return session?.streamsSortedByDetailedType()?.find { it.detailedType == StringConstants.pmTwoPointFive }
```
xameYeoh, line 109: So here I would have SensorName.PM2_5.detailedType from enum

app/src/main/java/pl/llp/aircasting/data/api/util/StringConstants.kt
xameYeoh: Another idea might be, if we want to keep all the constants in files, replace the strings in the enum with StringConstants.pmTwoPointFive so that we would have:
```
enum class SensorName(val detailedType: String) {
        F(StringConstants.F),
        PM(StringConstants.PM),
        // and so on
```
But in the defaultStream() method I would still use the enum value like SensorName.PM2_5.detailedType

xameYeoh (changes requested):
I've left some ideas on refactoring

app/src/main/java/pl/llp/aircasting/data/api/util/StringConstants.kt
LinX64: Didn't notice that enum class, thanks for mentioning it!

xameYeoh:
Looks super cool! 💪

PR #524, opened 2022-07-07
Binding expanded sessions

xameYeoh:
This includes Singleton for working with the Set of Expanded cards and saving it in the shared preferences; clearing the set on logout.
In the end, I went with saving/deleting them to SharedPreferences as soon as the card is expanded/collapsed, as you mentioned ;)
I thought that if some crash happens, we won't be able to save them, so better do it as soon as possible :D
And it does not seem to lag, so everything is fine

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/util/ExpandedCardsRepository.kt
+++ b/app/src/main/java/pl/llp/aircasting/util/ExpandedCardsRepository.kt
@@ -1,0 +1,35 @@
+package pl.llp.aircasting.util
+
+class ExpandedCardsRepository private constructor(
+    private val settings: Settings
+) {
+    private val uuids: MutableSet<String> = settings.getExpandedSessionsUUIDs()
+
+    companion object {
+        private var instance: ExpandedCardsRepository? = null
+        fun getInstance(): ExpandedCardsRepository? {
+            return instance
+        }
+
+        fun setup(settings: Settings) {
+            if (instance == null)
+                instance = ExpandedCardsRepository(settings)
+        }
+    }
+
+    fun add(uuid: String) {
+        uuids.add(uuid)
+        settings.saveExpandedSessionsUUIDs(uuids)
+    }
+
+    fun remove(uuid: String) {
+        uuids.remove(uuid)
+        settings.saveExpandedSessionsUUIDs(uuids)
+    }
+
+    fun clear() {
+        uuids.clear()
+    }
+
+    fun contains(uuid: String) = uuids.contains(uuid)
+}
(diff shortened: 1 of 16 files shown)


LinX64:
[Set saved session's expanded state when creating card with it](https://trello.com/c/nDlvr6D3/1602-set-saved-sessions-expanded-state-when-creating-card-with-it)

app/src/main/java/pl/llp/aircasting/util/ExpandedCardsRepository.kt
```diff
@@ -0,0 +1,35 @@
+package pl.llp.aircasting.util
+
+class ExpandedCardsRepository private constructor(
```
LinX64, line 3: Placing this Repository in the `util` package doesn't make sense. Either change the name to be a helper class or move it to the `local/repositories` as this is where it should be.

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/SessionManager.kt
```diff
@@ -105,6 +106,7 @@ class SessionManager(
 
     @Subscribe
     fun onMessageEvent(event: LogoutEvent) {
+        expandedCards()?.clear()
```
LinX64, line 109: `SharedPreferences as soon as the user clicks on the button` - Well, in that case, I would delete those saved `UUID`s (_along with other saved data on Preference_) as soon as the user hits the logout button on the settings.
Removing those saved `UUID`s from the preference in an event seems a bit weird and out of the current structure. Also, since it is a SharedPreference and you are saving it on the Settings.kt class, it should be deleted from the Settings and relavant classes, this way, it will make it too complicated and doesn't seem to be right to be honest. What do you think?

xameYeoh: I don't understand your point. The deletion of all the values from the shared preferences file (except BACKEND_URL_KEY and BACKEND_PORT_KEY), including our saved UUIDs, already takes place in the app, it has always been there:
```
private fun deleteFromSettings() {
        val keys = sharedPreferences.all.keys
        val editor = sharedPreferences.edit()
        for (key in keys) {
            if (key != BACKEND_URL_KEY && key != BACKEND_PORT_KEY) {
                editor.remove(key)
                editor.apply()
            }
        }
    }
```
This method gets fired from the Settings logout() method:
```
    open fun logout() {
        deleteFromSettings()
    }
```
And this logout method gets fired from the LogoutService: 
```
class LogoutService(private val mSettings: Settings) {
    fun perform(callback: (() -> Unit)? = null) {
        // to make sure downloading sessions stopped before we start deleting them
        EventBus.getDefault().post(LogoutEvent())
        **mSettings.logout()**
        clearDatabase()
        callback?.invoke()
    }
```
Which is fired from: 
```
override fun onSignOutClicked() {
        **logoutService.perform** {
            LoginActivity.startAfterSignOut(mContext)
        }
    }
```
When the Logout button is clicked

xameYeoh: Settings class does not hold any reference to the Set right now, this is still just a wrapper to access the the shared preferences editor methods and keys

app/src/main/java/pl/llp/aircasting/util/ExpandedCardsRepository.kt
xameYeoh: Good catch, forgot to move it :)

LinX64 (approved):
Looks good!

PR #526, opened 2022-07-08
S&F > bottomsheet: map, chart, buttons not displayed properly

LinX64:


LinX64:
We'll need to test this on a smaller screen.

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
```diff
@@ -176,17 +176,25 @@ class SearchFixedResultActivity : AppCompatActivity(), OnMapReadyCallback,
...
     }
 
+    private fun showLoader() {
+        AnimatedLoader(binding.searchLoader).start()
+    }
+
+    private fun stopLoader() {
+        AnimatedLoader(binding.searchLoader).stop()
+    }
+
```
xameYeoh, line 194: Does it create a new AnimatedLoader object each time these methods are called? I would have it as a private field to bind only once when the activity is created and call mLoader.start() or stop(). But maybe I am too picky :)

xameYeoh: I've just thought that maybe it would be nice instead of having this class and bind it to images, just create an extension from ImageView class and have startAnimation() and stopAnimation() extensions from it, where we would have for example:
```
fun ImageView.startAnimation() {
        visible()
        animatable = drawable as? Animatable
        animatable?.start()
    }
```
And another one like this for stopAnimation()

xameYeoh (commented):
I've left an idea on refactoring Loader

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedResultActivity.kt
LinX64: You're right, they've been using this animate for the other parts of the app, so having an ext func is a perfect idea, thanks!

app/src/main/java/pl/llp/aircasting/util/extensions.kt
```diff
@@ -235,3 +236,17 @@ fun GoogleMap.setMapType(mSettings: Settings, mContext: Context) {
     if (mSettings.isUsingSatelliteView()) this.setMapTypeToSatellite()
     else this.setMapTypeToNormalWithStyle(mContext)
 }
+
+fun ImageView.animateLoader(): Animatable {
+    return drawable as Animatable
+}
```
xameYeoh, line 242: I would name it animatable() if it is possible, as it does not animateLoader() but just returns drawable as animatable

xameYeoh (approved):
Naming comment, other than that looks cool! 😎  👍

PR #527, opened 2022-07-10
ProGuard | Recording fix session

LinX64:


LinX64:
[ProGuard | Recording fix session](https://trello.com/c/zyh9c5nw/1615-proguard-recording-fix-session)

PR #532, opened 2022-07-12
Duplicate new sessions

xameYeoh:


LinX64 (approved):
Interesting :) Let's go for test.

LinX64:
I can see that there is a start method in the MainActivity.kt as a companion object where they were using it for starting the MainActivity with the same flags that you mentioned:

```
companion object {
        fun start(context: Context?) {
            context?.let {
                val intent = Intent(it, MainActivity::class.java)
                intent.flags =
                    Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK or Intent.FLAG_ACTIVITY_CLEAR_TOP
                it.startActivity(intent)
            }
        }
    }
```

So maybe we can use the same method for creating the MainActivity?

PR #539, opened 2022-07-12
Leaked GCP API keys

LinX64:
New implementation based on the best practices by Google for hiding the important keys.


LinX64:
This card is related to the ProGuard - We still need to enable the obfuscation for the app.

xameYeoh (approved):

PR #540, opened 2022-07-13
ProGuard roles | Reduce apk size

LinX64:
This enables progaurd + shrinking resources, also reduces the size of the apk for release mode.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/util/helpers/sensor/SessionManager.kt
+++ b/app/src/main/java/pl/llp/aircasting/util/helpers/sensor/SessionManager.kt
@@ -154,11 +154,11 @@ class SessionManager(
         periodicallySyncSessionsService.pause()
     }
 
-    fun registerToEventBus() {
+    private fun registerToEventBus() {
         EventBus.getDefault().safeRegister(this)
     }
 
-    fun unregisterFromEventBus() {
+    private fun unregisterFromEventBus() {
         EventBus.getDefault().unregister(this)
     }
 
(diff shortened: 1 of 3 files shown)


LinX64:
[ProGuard | Recording fix session](https://trello.com/c/zyh9c5nw/1615-proguard-recording-fix-session)

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/SessionManager.kt
```diff
@@ -154,11 +154,11 @@ class SessionManager(
...
 
-    private fun registerToEventBus() {
+    fun registerToEventBus() {
         EventBus.getDefault().safeRegister(this)
     }
 
-    private fun unregisterFromEventBus() {
+    fun unregisterFromEventBus() {
         EventBus.getDefault().unregister(this)
     }
```
xameYeoh, line 163: I think these methods were called from the class itself only, or are they being called from somewhere else? If no, they can be private I guess

xameYeoh (commented):
Left comments

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/SessionManager.kt
LinX64: I need to check to see if this doesn't affect the `EventBus` with the ProGuard. but, good point! I'll look into it again!

xameYeoh (approved):

PR #541, opened 2022-07-13
Removed recursion

xameYeoh:


app/src/main/java/pl/llp/aircasting/ui/view/fragments/DashboardFragment.kt
```diff
@@ -88,12 +88,10 @@ class DashboardFragment : BaseFragment<DashboardViewMvcImpl, DashboardController
         requireActivity().let {
```
LinX64, line 88: Looks cool and interesting :) 

Anyways, I think we could write this logic in a better way with the safe call operator: (Let's call it refactoring)

 ```
activity?.onBackPressedDispatcher?.addCallback(viewLifecycleOwner) {
            if (isEnabled && mTabPosition == 0) {
                isEnabled = false
                showBatteryOptimizationDialogIfNeeded()
            }
        }
```

LinX64 (approved):
Interesting :)

PR #560, opened 2022-07-20
Search icon visible after going to let's begin

LinX64:


LinX64:
[Search icon visible after going to let's begin](https://trello.com/c/q44dhX2J/1656-search-icon-visible-after-going-to-lets-begin)

xameYeoh (approved):

PR #564, opened 2022-07-22
Rebinding updated sessions only | Full pr for refactoring recycler

xameYeoh:
This is a final pr for refactoring the recycler - it includes rebinding individually only those sessions that were changed/inserted/deleted when observer gets notified


app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
```diff
@@ -3,29 +3,37 @@ package pl.llp.aircasting.data.model.observers
...
 import pl.llp.aircasting.ui.view.screens.dashboard.SessionsViewMvc
-import kotlinx.coroutines.CoroutineScope
+import pl.llp.aircasting.ui.viewmodel.SessionsViewModel
 
 abstract class SessionsObserver<Type>(
     private val mLifecycleOwner: LifecycleOwner,
     private val mSessionsViewModel: SessionsViewModel,
     private val mViewMvc: SessionsViewMvc?
 ) {
+    enum class ModificationType {
```
LinX64, line 18: This enum class could go somewhere else. Let's separate it for the sake of Clean Architecture and Separation of Concerns.

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
```diff
@@ -3,29 +3,37 @@ package pl.llp.aircasting.data.model.observers
...
 
     private var mSessionsLiveData: LiveData<List<Type>>? = null
 
     private var mObserver: Observer<List<Type>> = Observer { dbSessions ->
         DatabaseProvider.runQuery { coroutineScope ->
             val sessions = dbSessions.map { dbSession -> buildSession(dbSession) }
             val sensorThresholds = getSensorThresholds(sessions)
-            if (anySessionChanged(sessions) || anySensorThresholdChanged(sensorThresholds)) {
-                onSessionsChanged(coroutineScope, sessions, sensorThresholds)
+            if (anySensorThresholdChanged(sensorThresholds)) updateSensorThresholds(sensorThresholds)
```
LinX64, line 34: Couldn't you use when in here?

```
when {
                anySensorThresholdChanged(sensorThresholds) -> updateSensorThresholds(
                    sensorThresholds
                )
                anySessionChanged(sessions) -> onSessionsChanged(coroutineScope, sessions)
            }

File: app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt [hunk 3]
```diff
@@ -50,24 +51,70 @@ abstract class SessionsRecyclerAdapter<ListenerType>(
...
+            }
+        }
+    }
+
+    private fun update(sessions: List<Session>?) {
+        sessions?.forEach { session ->
+            val position = mSessionUUIDS.indexOf(session.uuid)
+            if (found(position)) {
+                replaceSession(position, session)
+                val success = replacePresenter(session)
```
LinX64, line 79: Shouldn't we add the `success` variable beside the other variables? Like after the `position` val.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
```diff
@@ -80,59 +127,59 @@ abstract class SessionsRecyclerAdapter<ListenerType>(
...
+
+    private fun getFromDB(
+        session: Session,
+        dispatcher: CoroutineDispatcher = Dispatchers.IO
+    ): Session? = runBlocking {
+        withContext(dispatcher) {
+            val dbSessionWithMeasurements =
+                mSessionsViewModel.reloadSessionWithMeasurements(session.uuid)
+            return@withContext dbSessionWithMeasurements?.let {
+                Session(dbSessionWithMeasurements)
```
LinX64, line 181: We're using `let` to avoid nullability of the `dbSessionWithMeasurements` variable and then again you're passing the same variable to the `Session` inside the `let` scope. Don't make the same mistake as they did. Either pass `it`, or set a name for the lambda, and then pass it to `Session`;

```
 return@withContext dbSessionWithMeasurements?.let { Session(it) }
```

And/or:

```
 return@withContext dbSessionWithMeasurements?.let { dbSession ->
                Session(dbSession)
            }
```

app/src/main/java/pl/llp/aircasting/ui/view/screens/new_session/confirmation/ConfirmationController.kt
```diff
@@ -27,7 +26,7 @@ class ConfirmationController(
     }
 
     fun onStart(context: Context?) {
-        context?.hideKeyboard()
+        //context?.hideKeyboard()
```
LinX64, line 29: I think I reverted this back. So maybe you can use it as well because the relevant ext function should be already there.

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/measurement_table_container/MeasurementsTableContainer.kt
```diff
@@ -69,22 +70,20 @@ abstract class MeasurementsTableContainer {
         mSelectable = true
         mDisplayValues = displayValues
         if (displayValues) mMeasurementValues = mRootView?.measurement_values
-
-        refresh()
     }
 
-    fun makeCollapsed(displayValues: Boolean = true) {
+    fun makeCollapsed(sessionPresenter: SessionPresenter? = mSessionPresenter, displayValues: Boolean = true) {
```
LinX64, line 75: You are already passing the `mSessionPresenter` in the `SessionViewMvcImpl` where its  being called only from this class. Will we really need to assign the `mSessionPresenter` in this class too?

LinX64 (changes requested):
I left a few comments regarding some changes in this PR. Some of them are important to look and some of them are not.

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
xameYeoh: As far as I got it, when does not execute sequentially like in Java, so only one command will execute in this case. And these need to be executed both if both conditions are true

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
xameYeoh: Not sure what you mean here, like that? :
```
val position = mSessionUUIDS.indexOf(session.uuid)
val success = replacePresenter(session)
            if (found(position)) {
                replaceSession(position, session)
```
The replacing of presenter only makes sense if the UUID was found at the given position

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
xameYeoh: If this is still nullable in the lambda, shouldn't the compiler notify in such case?

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
LinX64: I see. "if both conditions are true" - then why not `if(firstCondition && second)`? Could be written simpler right?

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
xameYeoh: I have thought about whether to put it and how to name it for a long time, and after some research, I've come to a conclusion that enums should belong to the place where they are used. In these case, the enum is utilised only in this class, so I figured that there would be no point to transfer it to some other place. Besides, when its name is accessed like: SessionsObserver.ModificationType to me it sounds logical and full, - so you can first guess that it is tied to the Observer and it describes the Modification Type that was made on a Session. 
If this will be transferred somewhere else, I think that all these connections will somehow be needed to be put in the name, and here it is achieved by putting it in the only class where it is used

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
xameYeoh: No, not like that. Those are two separate events - when the thresholds change and when the session's data change. They happen irrespectively from each other.

app/src/main/java/pl/llp/aircasting/data/model/observers/SessionsObserver.kt
LinX64: This is true only if we'll be using this enum in this class and nowhere else. In this case, it makes sense to keep it there!

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
LinX64: I'm talking about this line: `val success = replacePresenter(session)` - I was wondering if it is possible to add this variable right after the first one one. If not, adding a space after this line `replaceSession(position, session)` will make it more understandable, what do you think?

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
LinX64: That was weird to me too! Not sure why its not notifying us but, the structure is pretty simple.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionsRecyclerAdapter.kt
xameYeoh: Like that?
```
            val success: Boolean
            val position = mSessionUUIDS.indexOf(session.uuid)
            if (found(position)) {
                replaceSession(position, session)
                success = replacePresenter(session)
                if (success) notifyItemChanged(position)
            }
```

LinX64 (approved):

xameYeoh:
The changes were transferred [here](https://github.com/HabitatMap/AircastingAndroid/pull/573) with better git history

PR #572, opened 2022-07-27
Zoom level limit

LinX64:


xameYeoh (commented):
I think this will zoom in when the camera stops, so you can theoretically still zoom out past the limit of 6f. Is it not possible in the sdk to actually limit the zoom so that we can not even zoom out of the the given zoom level (frame)?

LinX64:
> I think this will zoom in when the camera stops, so you can theoretically still zoom out past the limit of 6f. Is it not possible in the sdk to actually limit the zoom so that we can not even zoom out of the the given zoom level (frame)?

I tried to implement it like what they did on IOS. Instead found a better solution without zooming to that specific location. This should be a better solution for that!

xameYeoh (approved):

PR #576, opened 2022-07-29
Search and follow | New implementation

LinX64:


LinX64:
[Search and follow | New implementation](https://trello.com/c/8A5nQOLQ/1684-search-and-follow-new-implementation)

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionActivity.kt
```diff
@@ -0,0 +1,57 @@
...
+            rootActivity.startActivity(intent)
+        }
+    }
+
+    @Inject
+    lateinit var viewModelFactory: ViewModelProvider.Factory
+
+    @Inject
+    lateinit var searchFollowViewModel: SearchFollowViewModel
+
```
xameYeoh, line 32: I guess I kinda missed the point when this factory approach was added. I am not sure, but can we not access the viewmodel just like that?:
`val viewModel: SearchFollowViewModel by viewModels()`

app/src/main/java/pl/llp/aircasting/ui/view/fragments/SearchFollowResultFragment.kt
```diff
@@ -69,64 +68,75 @@ class SearchFixedResultActivity : BaseActivity(), OnMapReadyCallback,
...
-        mLat = intent.getStringExtra("lat").toString()
-        mLng = intent.getStringExtra("lng").toString()
+            mLat = it.getString("lat").toString()
+            mLng = it.getString("lng").toString()
 
-        txtParameter = intent.getStringExtra("txtParameter")
-        txtSensor = intent.getStringExtra("txtSensor")
+            txtParameter = it.getString("txtParameter")
+            txtSensor = it.getString("txtSensor")
+        }
```
xameYeoh, line 130: I guess if we already want to use fragments, it would be better to not use the arguments at all and pass the data between them using ViewModels. But I understand that it will need further work here, so maybe for now let's just rename it to getArgumentsFromLocationFragment()?

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionActivity.kt
LinX64: Short answer: No! Because we're assigning the viewModel again in the `setupFactory()` method and by `viewModels()` cannot serve as a delegate for `var`.

I didn't want to touch this, but I have a better solution for it, check my new commit for it.

app/src/main/java/pl/llp/aircasting/ui/view/fragments/SearchFollowLocationFragment.kt
```diff
@@ -1,63 +1,60 @@
...
-import pl.llp.aircasting.ui.view.common.BaseActivity
+import pl.llp.aircasting.databinding.FragmentSearchFollowLocationBinding
 import pl.llp.aircasting.util.Settings
 import pl.llp.aircasting.util.extensions.gone
 import pl.llp.aircasting.util.extensions.initializePlacesApi
 import pl.llp.aircasting.util.extensions.setStyle
 import pl.llp.aircasting.util.extensions.visible
 
-class SearchFixedSessionsActivity : BaseActivity() {
+class SearchFollowLocationFragment : Fragment() {
```
xameYeoh, line 29: I would have an inner directory SearchFollowFixed here, and name fragments like:
1. SearchLocationFragment
2. MapResultFragment

xameYeoh (commented):
Comments

app/src/main/java/pl/llp/aircasting/ui/view/fragments/SearchFollowResultFragment.kt
LinX64: I didn't want to change the implementation we had, and yes, it will need a refactor with [SharedViewModel](https://developer.android.com/topic/libraries/architecture/viewmodel#sharing) for sure.

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionActivity.kt
xameYeoh: But why do we need to re assign the viewModel and this factory? Could we not have the SearchFollowViewModel as `val` , get it in activity with `by viewModels()` once, and in fragments we already have:
 `val searchFollowViewModel by activityViewModels<SearchFollowViewModel>()`
 which grabs the same instance of ViewModel from the activity

xameYeoh: Okay, I guess I got it - because all the dependencies there are coming through dagger, and the factory along with the activity need to be constructed there as well

LinX64: There are two concerns here;

1. I just realized that I added a module for `ViewModels` in `ViewModelModule` to use injection with the dagger for ViewModels, that's why we were injecting the `ViewModel` in that way.

2. If we do it your way, what will happen if we need to use this viewModel for observing something else in the `Activity`?

xameYeoh (approved):

PR #577, opened 2022-07-29
No sessions displayed west of specific longitude

LinX64:


LinX64:
[No sessions displayed west of specific longitude](https://trello.com/c/GyHf8lcW/1682-no-sessions-displayed-west-of-specific-longitude)

app/src/main/java/pl/llp/aircasting/util/CalendarUtils.kt
```diff
@@ -1,14 +1,58 @@
...
+        private val mCalendar = calendar()
 
         fun dayOfMonth(date: Date): Int {
             mCalendar.time = date
             return mCalendar[Calendar.DAY_OF_MONTH]
         }
+
+        private fun getCurrentDayMilliSeconds(): Long {
+            return mCalendar.timeInMillis
+        }
```
xameYeoh, line 17: This method does not seem to be used

app/src/main/java/pl/llp/aircasting/util/CalendarUtils.kt
```diff
@@ -1,14 +1,58 @@
...
+
+        /**
+         * This method returns the end of the current day in milliseconds.
+         * timeZone is UTC.
+         * example: if the current day is Aug 1, the method returns:
+         * GMT: Monday, August 1, 2022 11:59:59 PM
+         * Our time zone: Tuesday, August 2, 2022 1:59:59 AM GMT+02:00 DST
+         **/
+        fun getTimeTo(): Long {
+            val today = calendar()
```
xameYeoh, line 47: I must say it looks very good with using calendar and it's very clear what the methods are doing when you read through them!
Some of my thought are:
1. If these methods are only used in the repository class, why would we want to extract them somewhere else? (If theoretically in future they will be needed elsewhere - this will be done at that time, but now, I think these methods are not general but an edge case which is only used for the S&F repository - so I would keep them as private of the class)
2. I would name them more clearly so that we could eliminate the large comments around them, how about:
`getStartOfTodayMillisFromYearAgo()` and `getEndOfTodayMillis()`?
With namings like that + clear calendar codes provide sufficient understanding of what methods are doing. 
And in Repository class there could be a simple comment saying:
"We grab sessions in time period starting from today's day a year ago till today"

xameYeoh (commented):
I've added some of my thoughts on structure

app/src/main/java/pl/llp/aircasting/util/CalendarUtils.kt
xameYeoh: And I would squash some commits as well to keep it short

LinX64: 1. In this case, it does not matter if we're gonna use it somewhere in the future or not, the repository shouldn't hold the logics and operations like working with Calender; please read [here](https://digital-solutions.consulting/uncategorized/repository-in-androids-mvvm-architecture/).

P.S: This is obviously an edge-case, however, the purpose of MVVM design pattern and Clean Architecture was the Separation of concerns, means that each class should be responsible for its job. Having two methods which is working with Calender inside the repository, doesn't make sense.

2. I agree, this naming is much better!

xameYeoh: If you want to lift this responsibility from the repository class, then I think you need to provide those values from other class to it through its method. In my view, right now the repository calls these methods by itself and I think it makes it the same as if they were already inside of it. But then it would not make a lot of sense as we only use it with this same logic
I do not see this as a violation of the repository's purpose. It provides the data in the format that is needed and it acts as a bridge - right according to the principle. It does the job it is being asked to do

xameYeoh: I will not push it any further, so if you want to keep it the way it is - it's fine and can be rebased

xameYeoh (approved):

xameYeoh (approved):

PR #578, opened 2022-07-31
Swapped java annotations with kotlin

xameYeoh:


app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedSessionActivity.kt
```diff
@@ -27,7 +27,6 @@ class SearchFixedSessionActivity : BaseActivity() {
     @Inject
     lateinit var viewModelFactory: ViewModelProvider.Factory
 
-    @Inject
```
LinX64, line 30: Same topic; https://github.com/HabitatMap/AircastingAndroid/pull/576#issuecomment-1199098414

If you remove the `@Inject` annotation, you'll no longer need to keep the `ViewModel` module in DI, and especially `bindSearchFollowViewModel` function. Maybe this was removed by mistake?

LinX64 (changes requested):
We already discussed this topic on another thread, I wrote a comment.

LinX64 (approved):

PR #581, opened 2022-08-01
DI for map fragment

xameYeoh:
This introduces DI for MapResultFragment with a factory for fragments (same approach we have with ViewModelFactory). This initially was needed to provide a SearchFollowViewModel in tests to the MapResultFragment

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/fragments/search_follow_fixed_session/SearchLocationFragment.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/fragments/search_follow_fixed_session/SearchLocationFragment.kt
@@ -7,7 +7,9 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.EditText
 import android.widget.ImageButton
+import androidx.core.os.bundleOf
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentFactory
 import com.google.android.gms.common.api.Status
 import com.google.android.libraries.places.api.Places
 import com.google.android.libraries.places.api.model.Place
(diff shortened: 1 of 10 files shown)


app/src/main/java/pl/llp/aircasting/di/factories/InjectFragmentFactory.kt
```diff
@@ -0,0 +1,24 @@
+package pl.llp.aircasting.di.factories
+
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentFactory
+import javax.inject.Provider
+
+class InjectFragmentFactory(
```
LinX64, line 7: 1. The `Inject` naming in here is not good. Let's just have it as `FragmentFactory` because of clean coding and the intention of this class.
2. Since you added this class inside the factories package, how about adding the other factory that we have from viewModel package to have them all in one place?
3. You provided the factory in the `AppModule`, and here there is no `Inject` in the constructor, which makes it different from the previous implementation. If you are trying to add another factory, it should be the same structure as we have or, probably, changing the old one to a new one to have a unified version of codes in an application.

app/src/main/java/pl/llp/aircasting/ui/view/fragments/search_follow_fixed_session/SearchLocationFragment.kt
```diff
@@ -33,7 +37,12 @@ class SearchLocationFragment : Fragment() {
     private var placesClient: PlacesClient? = null
     private var txtSelectedParameter: String = StringConstants.measurementTypePM
     private var txtSelectedSensor: String = StringConstants.openAQsensorNamePM
-    private val mSettings: Settings by lazy { Settings(requireActivity().application) }
+
+    @Inject
+    lateinit var mSettings: Settings
+
+    @Inject
+    lateinit var fragmentFactory: FragmentFactory
```
LinX64, line 45: And now here you are injecting the Fragment, just like the previous version of `ViewModelFactory`

app/src/main/java/pl/llp/aircasting/di/factories/InjectFragmentFactory.kt
xameYeoh: 1. Then there will be a conflict with FragmentFactory from android packages
2. Good point!
3. I have tried to make it same as it was with ViewModelFactory, but apparently dagger does not see this binding if I provide it with constructor injection (maybe I missed something, but I was not able to do it like that).

xameYeoh:
I've lifted the fragment reference from adapter completely in the last commit

LinX64 (approved):

PR #586, opened 2022-08-02
Circleci editor/451/fix tests espresso

LinX64:


LinX64:
Had to create another PR because of the CircleCI, which seems to be not updating the roles so I had to create a new PR from console.

PR #589, opened 2022-08-02
Search and follow Bugs

LinX64:


LinX64:
[Map screen: when pressing back from top toolbar, search screen does not retain previously searched place](https://trello.com/c/zJK0Xyv3/1686-map-screen-when-pressing-back-from-top-toolbar-search-screen-does-not-retain-previously-searched-place)

xameYeoh (changes requested):
Still not working on my side - goes to the dashboard

LinX64:
> Still not working on my side - goes to the dashboard

Can't test this because of my IP and restrictions. I'll fix it from Home.

LinX64:
It should work as expected now, for both navigation and data saving when replacing another fragment.

xameYeoh (approved):
Cool!

PR #590, opened 2022-08-02
Display empty string when the selected stream is null

xameYeoh:
This should eliminate the problem with 1hr average string jumping up when there are no streams and saying "1 hr avg - null" on UI

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/Chart.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/Chart.kt
@@ -12,6 +12,7 @@ import pl.llp.aircasting.R
 import pl.llp.aircasting.ui.view.screens.dashboard.SessionPresenter
 import pl.llp.aircasting.util.MeasurementColor
 import pl.llp.aircasting.util.TemperatureConverter
+import pl.llp.aircasting.util.extensions.inVisible
 
 class Chart(
     context: Context,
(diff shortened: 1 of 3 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/Chart.kt
```diff
@@ -114,7 +114,10 @@ class Chart(
     }
 
     private fun chartUnitText(): String {
-        return "${mContext.getString(chartUnitLabelId())} - ${mSessionPresenter?.selectedStream?.detailedType}"
+        return if (mSessionPresenter?.selectedStream == null)
+            mContext.getString(R.string.empty_string)
```
LinX64, line 118: There is no point in having a null string in the application. Instead of doing so, check if the selectedStream is null, then just hide the view.

LinX64 (changes requested):
I've added a comment, there is no point of having a null string.

app/src/main/res/values/strings.xml
```diff
@@ -522,4 +522,7 @@
     <string name="use_celsius_scale">Use Celsius scale</string>
     <string name="from">from</string>
     <string name="to">to</string>
+
+    <!-- Miscellaneous -->
+    <string name="empty_string" translatable="false" />
```
LinX64, line 527: This empty string still exist..

LinX64 (approved):

PR #591, opened 2022-08-02
Overlap icon | Refresh chart and measurement table unconditionally 

xameYeoh:
This should fix session card getting data from other session.
The problem consisted of two parts: when session did not have measurements, its measurements table was not bound and the chart was not bound either. So somehow, both table and chart were instead bound to other session presenter and displayed other session's data (maybe it was reused by adapter).
Now their instances are created and bound even if there are no measurements - they just don't display anything


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/charts/Chart.kt
```diff
@@ -60,8 +56,11 @@ class Chart(
         mLineChart?.invalidate()
     }
 
-    private fun prepareDataSet(): LineDataSet {
-        if (mEntries.isEmpty()) return LineDataSet(listOf(), "")
+    private fun prepareDataSet() {
+        if (mEntries.isEmpty()) {
+            mDataSet = LineDataSet(listOf(), "")
+            return
```
LinX64, line 62: Do we need really need this `return` here?

xameYeoh: So that we do not compute all the stuff that goes after this condition, as it will be useless if the entries are empty

LinX64 (approved):

PR #595, opened 2022-08-03
Tests Fix | API 21

LinX64:


LinX64:
[Fix tests | CircleCI](https://trello.com/c/PIfLvVyI/1693-fix-tests-circleci)

LinX64:
[Fix Tests | API 21](https://trello.com/c/xbXpLnl1/1697-fix-tests-api-21)

app/src/main/java/pl/llp/aircasting/ui/view/screens/create_account/CreateAccountActivity.kt
```diff
@@ -40,9 +41,15 @@ class CreateAccountActivity: BaseActivity() {
...
+        controller =
+            CreateAccountController(this, view, settings, apiServiceFactory, fromOnboarding)
 
         setContentView(view.rootView)
+
+        // Temporary fix for tests
+        // TODO: The requests should be done in a separate thread/maybe using MVVM for Login/register later
+        val policy = StrictMode.ThreadPolicy.Builder().permitAll().build()
+        StrictMode.setThreadPolicy(policy)
     }
```
xameYeoh, line 53: There is a class called `EspressoRunner` in the test directory which does the same. It should be utilised in AppConfig with `const val androidTestInstrumentation = "pl.llp.aircasting.helpers.EspressoRunner"` . Maybe it somehow gets missed in API 21?

xameYeoh: Because this error was in place for Room (which is strange because Room should crash immediately when you try to query on Main thread - so maybe it is an Espresso and Tests error only) as well when I was dealing with the tests for the first time, and using runner solved it everywhere (apparently not for 21 api for some reason)

xameYeoh: Also, I don't think this policy should be set in the actual production code - this should be scoped to tests only in my opinion

LinX64: That's correct, I wasn't also sure why this is happening, I need to investigate more on this for sure.
Anyways, as far as I understood, the problem was related to both Room `clearTable` and making the calls on the main thread.

I'll work on this more to find a better solution, I am still working on the last issue with a native crash, and in the meantime, I hope I'll be able to find a better solution except the policy for it.

xameYeoh (approved):

PR #596, opened 2022-08-03
Map | Bottomsheet screens additional tests and setup

xameYeoh:


app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
```diff
@@ -154,7 +191,133 @@ class SearchFollowTest {
...
+            )
+            .perform(click())
+
+        onView(withId(R.id.txtTitle))
+            .check(matches(textContainsString(cardTitle)))
+        onView(withId(R.id.txtDate))
+            .check(matches(textContainsString(cardDate)))
+        verifySwitchingChipsWork()
+
+        mapScenario.close()
```
LinX64, line 286: Looks good! 
However, it seems like I was missing something in the BottomSheet class. It means that if you run the test and lifeCycle reaches `onDestroyView()`, the test won't pass and will give you an error. So, the solution is to add `parentFragment` right after the `mapFragment`:
 
`        mapFragment?.parentFragment?.let { parentFragmentManager.beginTransaction().remove(it).commit() }`

Then test will pass and it will work as expected.

xameYeoh: Thanks for mentioning that! Got this error earlier today

app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
```diff
@@ -262,8 +262,11 @@ class SearchFollowTest {
     }
 
     @Test
-    fun whenChoosingCard_bottomSheetHasSameDateAndTitleAsCard_chipsSwitchGraphView() {
-        launchMapScreen(newYorkArgs)
+    fun whenChoosingCard_bottomSheetHasSameDateAndTitleAsCard_chipsSwitchGraphView_externalSessionIsFollowed() {
```
LinX64, line 265: This is more like a sentence rather than a function name. Shouldn't we make it shorter?

PR #598, opened 2022-08-08
BottomSheet Crash 

LinX64:
We were facing this crash when user wanted to re-open a session from the sessions list with bottomsheet. This was because of the `parentFragmentManager`, so instead of getting it like that, I used activity to get the `supportFragmentManager`.


LinX64:
[BottomSheet Crash](https://trello.com/c/khHpIiEl/1705-bottomsheet-crash)

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
```diff
@@ -265,6 +265,9 @@ class SearchFixedBottomSheet : BottomSheet(), OnMapReadyCallback {
     override fun onDestroyView() {
         super.onDestroyView()
 
-        mapFragment?.parentFragment?.let { parentFragmentManager.beginTransaction().remove(it).commit() }
+        mapFragment ?: return
+        mapFragment?.let {
+            activity?.supportFragmentManager?.beginTransaction()?.remove(it)?.commit()
```
czorek, line 270: Are all those things really nullable, even the transaction method?

czorek (approved):

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
LinX64: Good point! This was because of `activity` reference. I deleted it, and instead, it's using `requireActivity()` which seems more reasonable and removes the nullable part. Please check the last commit.

LinX64:
[Improvement for the RecyclerView Sessions list](https://trello.com/c/9kTSRcK3/1706-improvement-for-the-recyclerview-sessions-list)

app/src/main/java/pl/llp/aircasting/ui/view/screens/search/SearchFixedBottomSheet.kt
czorek: looks good to me :)

PR #601, opened 2022-08-10
AB: "PM2.5" should be the default parameter

LinX64:
So about the problem, I already set the stream to set pm2.5 as the default value, but for some reason, it didn't work for mobile active sessions, now it should be setting for all sessions including mobile active, but I remember that we were facing a problem by setting only to PM2.5, that's why this needs to be fully tested.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
@@ -28,7 +28,7 @@ class SessionPresenter() {
         loading: Boolean = false
     ) : this() {
         this.session = session
-        this.selectedStream = selectedStream ?: defaultPm2point5Stream(session)
+        this.selectedStream = selectedStream ?: defaultStream(session)
         this.expanded = expanded
         this.loading = loading
         this.sensorThresholds = sensorThresholds
@@ -65,7 +65,7 @@ class SessionPresenter() {
     }
 
     fun setDefaultStream() {
-        selectedStream = defaultPm2point5Stream(session)
+        selectedStream = defaultStream(session)
     }
 
     fun isFixed(): Boolean {
@@ -104,9 +104,13 @@ class SessionPresenter() {
     }
 
     companion object {
-        fun defaultPm2point5Stream(session: Session?): MeasurementStream? {
+        fun defaultStream(session: Session?): MeasurementStream? {
             val sortedByDetailedType = session?.streamsSortedByDetailedType()
-            return sortedByDetailedType?.find { it.detailedType == SensorName.PM2_5.detailedType }
+            val pm2point5 =
+                sortedByDetailedType?.find { it.detailedType == SensorName.PM2_5.detailedType }
+            val firstStream = sortedByDetailedType?.firstOrNull()
+
+            return pm2point5 ?: firstStream
         }
     }
 }
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
@@ -94,6 +94,10 @@ class SessionPresenter() {
         return session?.isAirBeam3() == true
     }
 
+    fun allStreamsHaveLoaded(): Boolean {
+        return session?.streams?.size == 5
+    }
+
     fun setSensorThresholds(sensorThresholds: List<SensorThreshold>) {
         val hash = hashMapOf<String, SensorThreshold>()
         sensorThresholds.forEach {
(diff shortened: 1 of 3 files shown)


LinX64:
[mobile active > AB: "PM2.5" should be the default parameter selection, currently it's "F"](https://trello.com/c/BNVf7ROD/1675-mobile-active-ab-pm25-should-be-the-default-parameter-selection-currently-its-f)

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -104,13 +104,9 @@ class SessionPresenter() {
...
     companion object {
-        fun defaultStream(session: Session?): MeasurementStream? {
+        fun defaultPm2point5Stream(session: Session?): MeasurementStream? {
             val sortedByDetailedType = session?.streamsSortedByDetailedType()
-            val pm2point5 =
-                sortedByDetailedType?.find { it.detailedType == SensorName.PM2_5.detailedType }
-            val firstStream = sortedByDetailedType?.firstOrNull()
-
-            return pm2point5 ?: firstStream
+            return sortedByDetailedType?.find { it.detailedType == SensorName.PM2_5.detailedType }
```
czorek, line 109: So what will happen now if the pm2.5 stream is not found? Do we have any fallback now? Do we need it?

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -28,7 +28,7 @@ class SessionPresenter() {
         loading: Boolean = false
     ) : this() {
         this.session = session
-        this.selectedStream = selectedStream ?: defaultStream(session)
+        this.selectedStream = selectedStream ?: defaultPm2point5Stream(session)
```
czorek, line 31: I think I might change the name to `defaultPm2_5Stream` to have it shorter and consistent with `SensorName.PM2_5`

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
LinX64: Good point!

I thought the same, I need to do more investigation on this.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
LinX64: Yes but lint says you cannot have an underscore in the function name, that was the reason why I added it like that.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -104,13 +104,19 @@ class SessionPresenter() {
...
+        fun defaultPm2point5Stream(session: Session?): MeasurementStream? {
             val sortedByDetailedType = session?.streamsSortedByDetailedType()
             val pm2point5 =
                 sortedByDetailedType?.find { it.detailedType == SensorName.PM2_5.detailedType }
-            val firstStream = sortedByDetailedType?.firstOrNull()
 
-            return pm2point5 ?: firstStream
+            return if (session != null && pm2point5 != null) pm2point5 else defaultFirstStream(
+                session
+            )
```
xameYeoh, line 119: Does not work, and to me it looks like it does the same as previous code but it does it with less clarity

LinX64: Please check the new approach with the new commit.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionViewMvcImpl.kt
```diff
@@ -172,9 +176,12 @@ abstract class SessionViewMvcImpl<ListenerType>(
     }
 
     private fun bindSelectedStream() {
-        if (mSessionPresenter != null && mSessionPresenter?.selectedStream == null) {
+        val session = mSessionPresenter?.session
+
+        if (session?.hasMeasurements() == true && session.mStreamsCount() > 5) {
             mSessionPresenter?.setDefaultStream()
         }
+
```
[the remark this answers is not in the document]
xameYeoh: There is no way it does not get called for external sessions - they are the same as local ones. This method gets called for all the session cards on all the tabs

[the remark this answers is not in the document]
xameYeoh: As for the Second thing when you call setDefaultStream yourself - I guess you refer to the external session being bound on the bottom sheet.   So you are saying that the default stream will be selected from there already?

PR #602, opened 2022-08-10
Missing 'hour range' and '1 min avg' on chart

LinX64:


LinX64:
[Missing 'hour range' and '1 min avg' on chart](https://trello.com/c/xkOB0Zvb/1707-missing-hour-range-and-1-min-avg-on-chart)

czorek (approved):

PR #604, opened 2022-08-10
Fix tests for API28-33

LinX64:
Most of the tests were failing because an external session was being followed and never unfollowed.
Also, I've added a few improvements to it; for ex; Running the tests based on the Names order.

### DIFF
--- a/app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
+++ b/app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
@@ -28,11 +28,9 @@ import org.hamcrest.CoreMatchers
 import org.hamcrest.Matcher
 import org.hamcrest.Matchers.*
 import org.hamcrest.core.AllOf.allOf
-import org.junit.After
-import org.junit.Before
-import org.junit.BeforeClass
-import org.junit.Test
+import org.junit.*
 import org.junit.runner.RunWith
+import org.junit.runners.MethodSorters
 import pl.llp.aircasting.data.api.services.ApiServiceFactory
 import pl.llp.aircasting.data.api.util.StringConstants.airbeam
 import pl.llp.aircasting.data.api.util.StringConstants.measurementTypeOzone
@@ -58,6 +56,7 @@ import javax.inject.Inject
 * Otherwise some test may fail
 * */
 
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 @RunWith(AndroidJUnit4::class)
 class SearchFollowTest {
     companion object {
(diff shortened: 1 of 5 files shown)


app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
```diff
@@ -424,13 +445,15 @@ class SearchFollowTest {
 
         onView(withId(R.id.places_autocomplete_search_bar))
             .perform(typeText(place))
+
+        Thread.sleep(3500)
+        // We have to wait for the list to be shown and to be retrieved from the server, otherwise tests will fail!
```
czorek, line 450: Can't we mock this somehow, do we really need to hit the server for data in tests? In general I think adding sleeps is usually not very reliable anyway

LinX64: Good point!
The thing is, we're loading data from Google Places API right after searching for a location. We cannot be sure how long it takes the data to be shown, so this was what I had in my mind.

Another approach is to check if the recyclerView is empty or not. I gave it a try and couldn't check if data is available or not, but I'll try to provide a check for it.

czorek: Sure, but isn't it possible to just not actually hit the external API but just mock the data?

LinX64: Sure we can, but as far as I understood, Kirill wrote these tests to have the same situation for different devices and to catch unexpected situations, but we can actually use dummy data for it, that's also a good idea.

czorek: Can we even check the behavior of different devices by running tests? How does that work?

LinX64: We can check most of the things on the Firebase Test lab video section and logs from the testings.

app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
```diff
@@ -290,26 +286,47 @@ class SearchFollowTest {
...
+        searchActivityScenario.close()
+    }
+
+    private fun unFollowFollowedSessionFromTheFollowingTab() {
+        goToMainScreen()
+
+        expandCard()
+
+        onView(withId(R.id.unfollow_button))
+            .perform(click())
```
xameYeoh, line 329: Discussed IRL: 
This is not related to unfollowing a session, as it only checks the state of the follow button **on the bottomsheet**, with the OpenAQ user logged in (so all the OpenAQ sessions should appear as your own)

LinX64:
I'll work on this and will check it again, so that we'll be able to merge it ASAP.

PR #607, opened 2022-08-12
Chart display unit symbol instead of detailed type

xameYeoh:


LinX64 (approved):
Looks good, let's test it.

PR #608, opened 2022-08-12
Ext session reorder wrong positions

xameYeoh:
This shares the functionality of following/unfollowing a session between SessionController and SearchFollowViewModel, which fixes the wrong external sessions' order while in reordering view


LinX64 (approved):
Good idea with sharing the functionalities👍

PR #610, opened 2022-08-12
Fix tests for API28-33

LinX64:


LinX64:
[Fix tests | 28-33 API new](https://trello.com/c/XGgdVXe5/1710-fix-tests-28-33-api-new)

LinX64:
[CircleCI | Update](https://trello.com/c/vVYlTbQc/1711-circleci-update)

LinX64:
[S&F > map: crash on session view](https://trello.com/c/EvxiAZNh/1712-sf-map-crash-on-session-view)

app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
```diff
@@ -287,29 +282,56 @@ class SearchFollowTest {
...
 
+    @Test
+    fun secondGoToFollowingTab_andUnfollow_the_followed_Session() {
+        mainActivityScenario = ActivityScenario.launch(startIntent)
+
+        unFollowFollowedSessionFromTheFollowingTab()
+
+        mainActivityScenario.close()
+    }
+
```
xameYeoh, line 293: If this is a test, I think there should be at least some sort of assertion and check going on here, it looks more like a helper function

xameYeoh (approved):

app/src/androidTest/java/pl/llp/aircasting/SearchFollowTest.kt
LinX64: That's right, but we either have to change the logic for the followed session tests or unfollow it to make it work. So I guess we can rely on this for now.

PR #612, opened 2022-08-16
Make size of labels fixed to prevent recalculation and wobbling of th…

xameYeoh:
…em when dragging slider

### DIFF
--- a/app/src/main/res/layout/hlu_slider.xml
+++ b/app/src/main/res/layout/hlu_slider.xml
@@ -22,7 +22,8 @@
         app:layout_constraintBottom_toTopOf="@id/track"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintTop_toTopOf="parent"
-        app:tint="@color/aircasting_grey_700" />
+        app:tint="@color/aircasting_grey_700"
+        tools:visibility="visible"/>
 
     <androidx.appcompat.widget.LinearLayoutCompat
         android:id="@+id/track"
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -616,5 +616,10 @@
         <item name="android:textColor">@color/aircasting_dark_blue</item>
     </style>
 
-
+    <style name="HLULabel" parent="Widget.AppCompat.TextView">
+        <item name="android:layout_width">25dp</item>
+        <item name="android:layout_height">20dp</item>
+        <item name="android:gravity">center</item>
+        <item name="android:visibility">invisible</item>
```
LinX64, line 623: Don't add the `visibility` and `gravity` attributes in the styles, this is not a good practice and is not easily maintainable.

app/src/main/res/layout/hlu_slider.xml
```diff
@@ -22,7 +22,8 @@
         app:layout_constraintBottom_toTopOf="@id/track"
```
LinX64, line 22: I just realized an extra `app:layout_constraintBottom_toBottomOf="parent"` in the `ConstraintLayout`, you may want to remove that as it has no effect.

app/src/main/res/values/styles.xml
xameYeoh: Why not? Otherwise I have to repeat it 3 times on the labels

LinX64: Okay, the better explanation for this would be to keeping most of the styles in the xml layout (as much as we can) because we have one style as file in the values and imagine what would happen if we add 5 more styles like this in there. So, the thing is that, when you add many styles in there, there will be a huge style file and maintaining it and/or editing it would be disaster on large-scale projects like this. As you can see, most of the Buttons and other widgets have different styles in there and it is already a huge file. That is why I mentioned that.

Anyways, unifying them is totally fine and is a great idea, but it shouldn't be done continuously because of the style file.

Also, I forgot to mention, the values that you used for the height and width should be defined in the dimensions (dimens.xml) as well.

PR #613, opened 2022-08-16
Check for all streams loaded only in local airbeam sessions to make m…

xameYeoh:
…ic sessions selected stream getting bound

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
@@ -95,7 +95,10 @@ class SessionPresenter() {
     }
 
     fun allStreamsHaveLoaded(): Boolean {
-        return session?.streams?.size == 5
+        if (session?.isAirBeam() == true && session?.isExternal == false)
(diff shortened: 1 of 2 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
```diff
@@ -95,7 +95,10 @@ class SessionPresenter() {
     }
 
     fun allStreamsHaveLoaded(): Boolean {
```
LinX64, line 97: How about this? 

```
fun allStreamsHaveLoaded(): Boolean {
        return if (session?.isAirBeam() == true && session?.isExternal == false) session?.streams?.size == 5 else true
    }
```

LinX64 (approved):
Could be written much simpler.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/SessionPresenter.kt
xameYeoh: I did it like that and had lint saying that in this structure "true" is unnecessary for some reason (don't know why)

PR #615, opened 2022-08-16
Follow a session > record mobile sound level session

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/screens/main/MainActivity.kt
```diff
@@ -60,24 +61,31 @@ class MainActivity : BaseActivity(), OnMapsSdkInitializedCallback {
...
         mNavController = navHostFragment.navController
 
         view?.setupNavController(mNavController)
-        view?.setupBottomNavigationBar(mNavController)
     }
 
     override fun onResume() {
         super.onResume()
         controller?.onResume()
+        view?.isMobileRecording()
```
xameYeoh, line 88: If all the other operations are done in the controller's onResume() method, can this call be transferred there as well?

app/src/main/java/pl/llp/aircasting/ui/view/screens/main/MainViewMvcImpl.kt
```diff
@@ -112,24 +118,39 @@ class MainViewMvcImpl(
...
+    fun isMobileRecording() {
+        DatabaseProvider.runQuery { scope ->
+            val isMobileActiveSessionExists = mSessionRepository.mobileActiveSessionExists()
+
+            DatabaseProvider.backToUIThread(scope) {
+                if (isMobileActiveSessionExists) goToMobileActiveTab()
+                else goToFollowingTab()
+            }
+        }
+    }
```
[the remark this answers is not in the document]
LinX64: I tried to add it in the controller, since we will need the reference of `mNavController` in the view, I had to use it that way. I remember that they were also using the same approach because NavController comes from the view and we can't risk it with using those extension functions because it's already has been initialised in the MainActivity, so i would put it there.

app/src/main/java/pl/llp/aircasting/ui/view/screens/main/MainActivity.kt
LinX64: I tried to add it in the controller, since we will need the reference of `mNavController` in the view, I had to use it that way. I remember that they were also using the same approach because NavController comes from the view and we can't risk it with using those extension functions because it's already has been initialised in the MainActivity, so i would put it there.

xameYeoh: This method is in MainViewMvcImpl, right? 
I mean that the MainController already has a reference to MainViewMvcImpl through an interface, so it can call its method by itself, if you make it available through interface

LinX64: Didn't see the interface reference, thanks for mentioning.

xameYeoh (approved):
Cool!

PR #617, opened 2022-08-17
Espresso test

LinX64:
This will solve the problem with increasing the size of the APK after releasing a build + the JDK problem with mismatch version.

### DIFF
--- a/buildSrc/src/main/java/Versions.kt
+++ b/buildSrc/src/main/java/Versions.kt
@@ -31,7 +31,7 @@ object Versions {
     const val openCv = "4.6"
 
     // Tests
-    const val espresso = "3.4.0"
+    const val espresso = "3.5.0-alpha07"
     const val mockito = "4.6.0"
     const val junit = "4.13.2"
     const val kotlinJunit = "1.6.20"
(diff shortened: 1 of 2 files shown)


LinX64:
[Espresso | APK Size Increase](https://trello.com/c/bEioLKyo/1718-espresso-apk-size-increase)

buildSrc/src/main/java/Versions.kt
```diff
@@ -31,15 +31,15 @@ object Versions {
     const val openCv = "4.6"
 
     // Tests
-    const val espresso = "3.4.0"
+    const val espresso = "3.5.0-alpha07"
```
xameYeoh, line 34: Do we really want to use the pre released version?

xameYeoh (approved):

buildSrc/src/main/java/Versions.kt
LinX64: We can't use 3.4.0 version because of the size problem, we can't also revert it back as we will face the problem with view and chips, so the only option which was recommended by an official is to use the alpha version and then when it was stable, using the latest version (3.5.0).

PR #618, opened 2022-08-18
Navigation | Refactor 2

LinX64:
This PR is related to the navigation with the Mobile Active Tab.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
           name: Test with Firebase Test Lab
           command: >
             sudo gcloud firebase test android run \
-              --device model=gts4lltevzw,version=28,locale=en,orientation=portrait \
+              --device model=ASUS_X00T_3,version=28,locale=en,orientation=portrait \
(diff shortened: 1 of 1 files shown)


LinX64:
[mobile dormant & fixed: when returning from map or graph, navigation takes you to the home tab rather than staying on mobile dormant or fixed](https://trello.com/c/U3OuRG9B/1719-mobile-dormant-fixed-when-returning-from-map-or-graph-navigation-takes-you-to-the-home-tab-rather-than-staying-on-mobile-dormant)

xameYeoh:
Is this included in [this pr](https://github.com/HabitatMap/AircastingAndroid/pull/621)? If so, I've rebased it already. I am not sure if the first commit from here will be duplicated or not, maybe we should remove it from here then?

LinX64:
I need to test the Firebase lab, after that I'll close this PR.

LinX64:
[Tests Refactor | New Navigation](https://trello.com/c/OieU6Rtt/1722-tests-refactor-new-navigation)

PR #619, opened 2022-08-18
Android 12 If the bluetooth is on when installing appb

LinX64:
I've also included the fix for tests in this PR.

### DIFF
--- a/app/src/androidTest/java/pl/llp/aircasting/FixedSessionTest.kt
+++ b/app/src/androidTest/java/pl/llp/aircasting/FixedSessionTest.kt
@@ -115,6 +115,7 @@ class FixedSessionTest {
         testRule.launchActivity(null)
 
         whenever(bluetoothManager.isBluetoothEnabled()).thenReturn(true)
+        whenever(permissionsManager.bluetoothPermissionsGranted(any())).thenReturn(true)
         whenever(permissionsManager.locationPermissionsGranted(any())).thenReturn(true)
         stubPairedDevice(bluetoothManager)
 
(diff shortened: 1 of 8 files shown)


LinX64:
[Android 12 If the bluetooth is on when installing app, it does not ask a permission for it and the phone can't see any airbeams](https://trello.com/c/FfYku2Ta/1651-android-12-if-the-bluetooth-is-on-when-installing-app-it-does-not-ask-a-permission-for-it-and-the-phone-cant-see-any-airbeams)

LinX64:
[Tests Refactor | New Navigation](https://trello.com/c/OieU6Rtt/1722-tests-refactor-new-navigation)

xameYeoh (commented):
Looks good!
I have tested it on my end with fixed sessions, and the outcome is that it asks for a permission, goes to the "choose device" screen (so everything appears to be working - it sees the airbeam), but when I try to connect to the airbeam, it is showing "connecting" screen indefinitely. If I go out of the wizard though, and come back - in connects to the airbeam without a problem. It seems like it actually connects to it first time as well, but the "connecting" screen does not get notified and thinks that we are still connecting

xameYeoh (commented):
Still same

LinX64:
> Still same

I know, I just rebased it with the master, still working on it.

LinX64:
> Looks good! I have tested it on my end with fixed sessions, and the outcome is that it asks for a permission, goes to the "choose device" screen (so everything appears to be working - it sees the airbeam), but when I try to connect to the airbeam, it is showing "connecting" screen indefinitely. If I go out of the wizard though, and come back - in connects to the airbeam without a problem. It seems like it actually connects to it first time as well, but the "connecting" screen does not get notified and thinks that we are still connecting

"It seems like it actually connects to it first time as well" - Regarding this sentence, this is not exactly true. Imagine what would happen to Android 12+ devices, where those devices need Bluetooth_SCAN and Bluetooth_CONNECT permission. You may try to remove all those permissions from your device and try this again, then you will see that it needs those permissions right after enabling the Bluetooth.

xameYeoh:
It looks like it connects, because the airbeam3 glows blue - which is supposedly meant that it was connected to. And if you go out of wizard (press back multiple times) and go back again to select the airbeam, it connects to it instantly (without asking any other permissions second time) - and all that happens on android 12 (so it seems like we do ask for both Scan and Connect permissions there)

xameYeoh (approved):
Works! 
Super!

PR #624, opened 2022-08-19
iOS like spinner behaviour

xameYeoh:
This includes spinner behaviour like in iOS - meaning that we will only have one spinner for Sync operation that will live across tabs on the dashboard, along with some major-ish refactor of the touched parts.

Some of the changes in structure that were made during the implementation process:
- Remove spinner handling from anywhere except Dashboard components (Mvc, Controller)
- Moving SwipeRefreshLayout from tabs (as there were 4 layout tabs programmatically created we had 4 Swipe layouts and 4 spinners) to Dashboard layout, so that we have ViewPager with 4 tabs inside a SwipeRefreshLayout, and swipe down on a ViewPager triggers the sync and spinner (when you are at the top of the list only). 
As the SwipeRefreshLayout appears right from the top of its child layout, it was placed in the Dashboard layout.
- Moving initial Sync operation when the user logs in from Main components (Controller) to Dashboard components. 
This was needed to be done as we want the spinner to be tied to initial sync operation on log in - and we will not able to notify it about sync getting started otherwise because at the moment Main controller is calling a sync operation the Dashboard (along with spinner) is not yet drawn, and its components are not yet instantiated - so they will not get notified about the initial sync. That's why the initial sync got transferred there - so the spinner operation is tied to it

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/fragments/DashboardFragment.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/fragments/DashboardFragment.kt
@@ -8,20 +8,30 @@ import androidx.activity.addCallback
 import com.google.android.material.tabs.TabLayout
 import kotlinx.android.synthetic.main.fragment_dashboard.*
 import pl.llp.aircasting.AircastingApplication
+import pl.llp.aircasting.data.api.services.ApiServiceFactory
+import pl.llp.aircasting.data.api.services.SessionsSyncService
 import pl.llp.aircasting.ui.view.common.BaseFragment
 import pl.llp.aircasting.ui.view.screens.dashboard.DashboardController
 import pl.llp.aircasting.ui.view.screens.dashboard.DashboardPagerAdapter
 import pl.llp.aircasting.ui.view.screens.dashboard.DashboardViewMvcImpl
-import pl.llp.aircasting.util.*
+import pl.llp.aircasting.util.Settings
+import pl.llp.aircasting.util.exceptions.ErrorHandler
 import pl.llp.aircasting.util.extensions.adjustMenuVisibility
 import pl.llp.aircasting.util.extensions.isIgnoringBatteryOptimizations
 import pl.llp.aircasting.util.extensions.showBatteryOptimizationHelperDialog
+import pl.llp.aircasting.util.isSDKGreaterOrEqualToM
 import javax.inject.Inject
 
 class DashboardFragment : BaseFragment<DashboardViewMvcImpl, DashboardController>() {
 
     @Inject
     lateinit var settings: Settings
+
+    @Inject
+    lateinit var errorHandler: ErrorHandler
+
+    @Inject
+    lateinit var apiServiceFactory: ApiServiceFactory
 
     private var mTabPosition: Int = 0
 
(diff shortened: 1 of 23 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/fragments/DashboardFragment.kt
```diff
@@ -40,7 +50,12 @@ class DashboardFragment : BaseFragment<DashboardViewMvcImpl, DashboardController
                 it, DashboardPagerAdapter.TABS_COUNT
             )
         }
-        controller = DashboardController(view)
+        val sessionsSyncService = SessionsSyncService.get(
+            apiServiceFactory.get(settings.getAuthToken()!!),
```
LinX64, line 54: Still you're using double-bang, maybe you can replace it with a scope function like `let` which is more safer?

app/src/main/res/layout/fragment_dashboard.xml
```diff
@@ -6,19 +6,29 @@
...
-            app:tabMode="scrollable" />
-    </androidx.viewpager.widget.ViewPager>
+    <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
+        android:id="@+id/refresh_sessions"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/tabs"
+        app:layout_constraintVertical_bias="0.479">
```
LinX64, line 26: This is extra. It's being used only for the preview.

app/src/main/res/layout/fragment_dashboard.xml
```diff
@@ -6,19 +6,29 @@
...
+        android:paddingEnd="@dimen/keyline_4"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:tabMode="scrollable" />
 
-        <com.google.android.material.tabs.TabLayout
-            android:id="@+id/tabs"
-            style="@style/Widget.Aircasting.TabLayout"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
```
LinX64, line 18: TabLayout doesn't have the start from the parent property. `(app:layout_constraintStart_toStartOf="parent")`

app/src/main/res/layout/fragment_dashboard.xml
```diff
@@ -6,19 +6,29 @@
...
-            android:paddingEnd="@dimen/keyline_4"
-            app:layout_constraintBottom_toTopOf="parent"
-            app:layout_constraintTop_toTopOf="parent"
-            app:tabMode="scrollable" />
-    </androidx.viewpager.widget.ViewPager>
+    <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
```
LinX64, line 19: `SwipeRefreshLayout` also doesn't have `app:layout_constraintEnd_toEndOf="parent"`.

LinX64 (approved):
I've left some comments, overall it's good.

app/src/main/res/layout/fragment_dashboard.xml
xameYeoh: Do you mean app:layout_constraintVertical_bias="0.479"?

LinX64: Ohh my bad didn't see the exact name. Yes, you don't need it actually because of         `app:layout_constraintBottom_toBottomOf="parent"`.

PR #626, opened 2022-08-21
Following > rearrange > clear: sessions aren't cleared

LinX64:


app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/helpers/SessionFollower.kt
```diff
@@ -22,43 +28,46 @@ class SessionFollower @Inject constructor(
...
     }
 
-    private fun updateFollowedAt(session: Session) {
-        DatabaseProvider.runQuery {
+    private fun updateFollowedAt(session: Session) = runBlocking {
+        launch(ioDispatcher) {
             mSessionRepository.updateFollowedAt(session)
             mSessionRepository.updateOrder(session.uuid, mSettings.getFollowedSessionsNumber())
         }
     }
```
xameYeoh, line 43: Doesn't it block the main thread now if you run runBlocking from UI?

xameYeoh: https://stackoverflow.com/questions/64768650/does-runblockingdispatchers-io-block-the-main-thread-even-the-the-coroutine-co

LinX64: Good question, please notice that I'm launching it with an `IO` dispatcher, along with `suspending` function. Please [read this](https://elizarov.medium.com/the-reason-to-avoid-globalscope-835337445abc).

Anyways, I can see that you were using the `runBlocking()` with `withContext()` in the SessionsRecyclerAdapter class, and you didn't use suspending function in there. We should ad the suspending function to it later.

xameYeoh: > Anyways, I can see that you were using the `runBlocking()` with `withContext()` in the SessionsRecyclerAdapter 
True, it was like that, maybe we should change it as well. But it was the only place where the runBlocking was in place across the production code, and judging by the documentation it seems that we will have the main thread blocked by all these operations now, even if you use the IO dispatcher. I understand that the usage of GlobalScope is out of place here, but as far as I get it, it does not block the ui when running these operations at least. Maybe we should gradually swap the places that use Global scope with the Structured concurrency that you mention, and later when it will be done we can safely remove it?

LinX64: I added another card regarding this, it needs a full refactor. I also used CoroutinesScope(), this was the best we could do about it, at least for now.

app/src/main/java/pl/llp/aircasting/ui/view/screens/dashboard/reordering_following/ReorderingFollowingRecyclerAdapter.kt
```diff
@@ -58,20 +79,18 @@ class ReorderingFollowingRecyclerAdapter (
...
         Collections.swap(mSessionUUIDS, firstPosition, secondPosition)
-        DatabaseProvider.runQuery {
-            mSessionsViewModel.updateOrder(mSessionUUIDS[secondPosition], secondPosition)
-            mSessionsViewModel.updateOrder(mSessionUUIDS[firstPosition], firstPosition)
+
+        runBlocking {
+            launch(Dispatchers.IO) {
+                mSessionsViewModel.updateOrder(mSessionUUIDS[secondPosition], secondPosition)
+                mSessionsViewModel.updateOrder(mSessionUUIDS[firstPosition], firstPosition)
+            }
```
xameYeoh, line 87: Do we still want to block here?

xameYeoh (approved):

LinX64:
[following > rearrange > clear: sessions aren't cleared](https://trello.com/c/s2a5eZwp/1720-following-rearrange-clear-sessions-arent-cleared)

PR #627, opened 2022-08-23
Record mobile session > exit app > enter app: app navigates 

LinX64:


LinX64:
[GlobalScope() Refactor](https://trello.com/c/DbJUfIVg/1725-globalscope-refactor)

LinX64:
[record mobile session > exit app > enter app: app navigates to the following tab rather then staying on the mobile active tab](https://trello.com/c/2j2flG6g/1714-record-mobile-session-exit-app-enter-app-app-navigates-to-the-following-tab-rather-then-staying-on-the-mobile-active-tab)

xameYeoh (commented):
I tested it and when I have active mobile session recording and try to navigate to following tab it goes immediately to mobile active (so I can't access following tab when I have an active session)

app/src/main/java/pl/llp/aircasting/ui/view/screens/main/MainViewMvcImpl.kt
```diff
@@ -122,8 +122,7 @@ class MainViewMvcImpl(
                 R.id.navigation_dashboard -> {
                     mSettings?.getFollowedSessionsNumber()
                         ?.let { rootActivity.adjustMenuVisibility(true, it) }
-
-                    navigateToAppropriateTab()
+                    rootActivity.goToFollowingTab()
```
xameYeoh, line 125: I think it might be better to swap places for these two, so that we adjust afterMenuVisibility after going to the following tab:
`
rootActivity.goToFollowingTab()
mSettings?.getFollowedSessionsNumber()
                        ?.let { rootActivity.adjustMenuVisibility(true, it) }
`
Because as far as I remember the `adjustMenuVisibility()` checks if we are on the following tab

xameYeoh (approved):
Cool

app/src/main/java/pl/llp/aircasting/ui/view/screens/main/MainViewMvcImpl.kt
xameYeoh, line 125: I think it might be better to swap places for these two, so that we adjustMenuVisibility after going to the following tab:
```
rootActivity.goToFollowingTab()
mSettings?.getFollowedSessionsNumber()
     ?.let { rootActivity.adjustMenuVisibility(true, it) }
```

Because as far as I remember the `adjustMenuVisibility()` checks if we are on the following tab

PR #634, opened 2022-08-24
Sync sessions on sign out

xameYeoh:
This does not include UI yet


xameYeoh:
Corresponding pr: https://github.com/HabitatMap/AircastingAndroid/pull/639

PR #636, opened 2022-08-25
Following > third party sessions > graph: midnight line not accurate

LinX64:


LinX64:
[Following > third party sessions > graph: midnight line not accurate](https://trello.com/c/roVpz9TK/1670-following-third-party-sessions-graph-midnight-line-not-accurate)

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
```diff
@@ -0,0 +1,25 @@
+package pl.llp.aircasting.util
+
+import java.util.*
+
+class DateUtil {
```
xameYeoh, line 5: I would name it differently because there is api DateUtils from apache that is used in the code as well (or maybe we can transfer all `DateUtils.truncate()` here as well?)

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        /**
+         * We need to subtract startTime because
+         * otherwise we lose precision while converting Long to Float
+         * and Float is needed for the MPAndroidChart library
+         **/
+        fun convertDateToFloat(date: Date): Float {
+            return (date.time - getStartTime()).toFloat()
+        }
```
xameYeoh, line 23: Is this code used elsewhere apart from GraphContaner? If not, I would not transfer it to another class. It seems like all the data for graph comes from the DataGenerator class, and I think it's better to keep everything related to data for graph there, so that we don't need to jump to some other classes while working only on the graph

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
LinX64: Renaming it is another thing, transferring the apache API there is another thing. I would do it in another separate PR (as a refactoring) 

For now, I renamed it to be more consistent.

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
LinX64: You're right but no, that's not the point!
Think of the future, when we will probably need those methods and we can get access to it easily through this helper class.

I know you're trying to keep everything in there, but since I moved everything to MVVM, we have to separate those classes based on their purposes and this will help the project to be more maintainable not just for now, but also **for the future.**

xameYeoh: If you want to have a general purpose class for date conversion, I think it is a good idea. In this case though, I would have methods that serve these general purposes. Meaning, that they should not serve only Graph's purposes (and now they do, because they are tied to use `getStartTime()` which is tied to the Graph only). 
If we want to follow this idea, I would have the method `convertDateToFloat()` to actually just convert date to float:
```
fun convertDateToFloat(date: Date): Float {
            return date.toFloat()
        }
```
or 
fun dateFromFloat(mFloatDate: Float): Date {
            return Date(mFloatDate.toLong())
        }
```
But now we could just add them as extensions already.

If we want them to not serve a general purpose, but instead serve only the graph's purpose - then I do not see a point of having them separately as it just plants confusion in my opinion (as if we are saying that they are used elsewhere, especially with their general names)

What do you think? If you still want to leave it as is, I will not push any further

xameYeoh (approved):

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
LinX64: There are many concerns regarding those methods and adding them inside those classes, but I will pass for now.

I have decided to remove the Util class and used them inside those classes, for now, please check the latest commit.

xameYeoh: We can come back to it and find a better code organisation together :)

PR #639, opened 2022-08-26
Sync on logout

xameYeoh:
This includes syncing on logout and waiting for the sync and logout operations (clearing of settings, database) to be done before being able to log back again. 
Some of the changes are:
- Further transfer of dependency management to Dagger (for `Settings`, `LogoutService`, `SessionSyncService`)
- Vector image change for loader 
- Change of events firing and listening in `SessionsSyncService`

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
@@ -75,8 +75,8 @@ class SessionsSyncService private constructor(
     }
 
     @Subscribe
-    fun onMessageEvent(event: LogoutEvent) {
-        sync()
+    fun onMessageEvent(logout: LogoutEvent) {
+        if (logout.inProgress) sync()
     }
 
     fun destroy() {
--- a/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
@@ -8,7 +8,6 @@ import pl.llp.aircasting.data.api.params.SyncSessionBody
 import pl.llp.aircasting.data.api.params.SyncSessionParams
 import pl.llp.aircasting.data.api.response.SyncResponse
 import pl.llp.aircasting.data.api.response.UploadSessionResponse
-import pl.llp.aircasting.data.local.DatabaseProvider
 import pl.llp.aircasting.data.local.repository.MeasurementStreamsRepository
 import pl.llp.aircasting.data.local.repository.NoteRepository
 import pl.llp.aircasting.data.local.repository.SessionsRepository
@@ -76,7 +75,8 @@ class SessionsSyncService private constructor(
 
     @Subscribe
     fun onMessageEvent(logout: LogoutEvent) {
-        if (logout.inProgress) sync()
+        if (logout.inProgress)
+            sync()
     }
 
     fun destroy() {
@@ -122,12 +122,13 @@ class SessionsSyncService private constructor(
                                 upload(body.upload)
                                 download(body.download)
                                 EventBus.getDefault().post(SessionsSyncSuccessEvent())
+                                syncStarted.set(false)
+                                EventBus.getDefault().postSticky(SessionsSyncEvent(false))
                             }
                         }
                     } else handleSyncError(shouldDisplayErrors, call)
 
-                    syncStarted.set(false)
-                    EventBus.getDefault().postSticky(SessionsSyncEvent(false))
+
                 }
 
                 override fun onFailure(call: Call<SyncResponse>, t: Throwable) {
@@ -205,7 +206,8 @@ class SessionsSyncService private constructor(
     ) {
         if (!call.isCanceled && !syncInBackground.get()) {
             EventBus.getDefault().post(SessionsSyncErrorEvent())
-
+            syncStarted.set(false)
+            EventBus.getDefault().postSticky(SessionsSyncEvent(false))
             if (shouldDisplayErrors) errorHandler.handleAndDisplay(SyncError(t)) else errorHandler.handle(
                 SyncError(t)
             )
--- a/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
@@ -32,21 +32,9 @@ class SessionsSyncService private constructor(
     private val errorHandler: ErrorHandler,
     private val settings: Settings
 ) {
-    private val uploadService: MobileSessionUploadService =
-        MobileSessionUploadService(apiService, errorHandler)
-    private val downloadService: SessionDownloadService =
-        SessionDownloadService(apiService, errorHandler)
-    private val removeOldMeasurementsService: RemoveOldMeasurementsService =
-        RemoveOldMeasurementsService()
-
-    private val sessionRepository = SessionsRepository()
-    private val measurementStreamsRepository = MeasurementStreamsRepository()
-    private val noteRepository = NoteRepository()
-    private val gson = Gson()
-    private val syncStarted = AtomicBoolean(false)
-    private var syncInBackground = AtomicBoolean(false)
-    private var triedToSyncBackground = AtomicBoolean(false)
-    private var mCall: Call<SyncResponse>? = null
+    init {
+        EventBus.getDefault().safeRegister(this)
+    }
 
     companion object {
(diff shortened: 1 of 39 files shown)


app/src/main/res/layout/activity_create_account.xml
```diff
@@ -200,15 +199,30 @@
...
             android:textAppearance="@style/TextAppearance.Aircasting.Login.TextButton"
-            app:layout_constraintBottom_toBottomOf="@id/login_prompt"
-            app:layout_constraintStart_toEndOf="@id/login_prompt"
-            app:layout_constraintTop_toBottomOf="@id/create_account_button"
+            app:layout_constraintBottom_toBottomOf="@+id/login_prompt"
+            app:layout_constraintStart_toEndOf="@+id/login_prompt"
+            app:layout_constraintTop_toTopOf="@+id/login_prompt"
             app:rippleColor="@null" />
 
+        <include
```
LinX64, line 211: The idea of including a `TextView` in here is not correct. It could be only `TextView` with the loader image in here. <Include is mostly being used for adding another complex layout inside another, but here it is just a `TextView` as I can see. 

Please [read](https://developer.android.com/develop/ui/views/layout/improving-layouts/reusing-layouts).

app/src/main/res/layout/logging_out_in_background.xml
```diff
@@ -0,0 +1,13 @@
...
+<TextView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/logout_events_in_progress"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_gravity="center"
+    android:text="@string/login_out_background"
+    android:textAppearance="?attr/textAppearanceHeadline4"
+    android:textSize="@dimen/text_size_s"
+    app:drawableLeftCompat="@drawable/ic_loader"
```
LinX64, line 11: **First:** "Compound drawables allowing the use of vector drawable when running on older versions
             of the platform."
             
**Second:** Is this going to be a `TextView` or `TextView/loader`? Is this a new implemented-element?

You could easily add the `loader` as an `ImageView` (just like what they did before) and add a `TextView` after it inside a `ConstraintLayout` or `LinearLayout` (with `Horizontal` orientation)

xameYeoh: That's exactly what I did first time, and that's why it was an <include> layout before. This is going to be a textview+drawable, so android studio gave me a hint to better use just one TextView with compound drawable in it instead of LinearLayout which will have ImageView and TextView

app/src/main/res/layout/activity_create_account.xml
xameYeoh: This was going to be a linear layout with two elements. I would still keep the include though, as this is a repeating element (it is used on two screens) so that I don't need to copy paste it. But if you think that it's crucial, we can change that

app/src/main/res/layout/logging_out_in_background.xml
LinX64: It's just s suggestion, you don't have to use.

Anyways, How about this?

        <include
            layout="@layout/logging_out_in_background"
            android:layout_marginStart="@dimen/keyline_4"
            android:layout_marginTop="10dp"
            android:layout_marginEnd="@dimen/keyline_4"
            android:visibility="invisible"
            app:layout_constraintBottom_toBottomOf="parent"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toBottomOf="@+id/login_prompt"
            tools:visibility="visible" />
            
And the layout: 

```
<?xml version="1.0" encoding="utf-8"?>
<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_height="wrap_content">

    <ImageView
        android:id="@+id/loader_image"
        android:layout_width="wrap_content"
        android:src="@drawable/ic_loader"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:tint="@color/aircasting_blue_400"
        tools:visibility="visible" />

    <TextView
        android:id="@+id/txt_logout_background"
        android:layout_width="0dp"
        android:layout_marginStart="@dimen/keyline_2"
        android:maxLines="3"
        android:text="@string/login_out_background"
        android:textAppearance="?attr/textAppearanceHeadline4"
        android:textSize="@dimen/text_size_s"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toEndOf="@+id/loader_image"
        app:layout_constraintTop_toTopOf="parent" />

</androidx.constraintlayout.widget.ConstraintLayout>
```

Like I mentioned, `<include` tag is mostly being used for including another layout inside another, and it's not good to include only one TextView in there. I agree that compound view might seem to be easy and easily definable, but separating those elements inside another layout will give us more richer control over those elements, and I was also worried about `drawableLeftCompat`.

app/src/main/java/pl/llp/aircasting/util/Settings.kt
```diff
@@ -2,52 +2,52 @@ package pl.llp.aircasting.util
 
 import android.app.Application
```
LinX64, line 3: Settings class will need an import optimize.

app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
```diff
@@ -109,12 +121,13 @@ class SessionsSyncService private constructor(
                                 upload(body.upload)
                                 download(body.download)
                                 EventBus.getDefault().post(SessionsSyncSuccessEvent())
+                                syncStarted.set(false)
+                                EventBus.getDefault().postSticky(SessionsSyncEvent(false))
```
LinX64, line 125: Maybe you could have two private methods for `SessionSyncSuccess` and `SessionSyncFailed`?

app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
```diff
@@ -66,6 +69,16 @@ class SessionsSyncService private constructor(
         }
     }
```
LinX64, line 71: Couldn't we `@Inject` those parameters inside the SessionSyncService class constructor instead of keeping them private?

Sync they are all provided in dagger, we should be able to inject them.

app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
```diff
@@ -66,6 +69,16 @@ class SessionsSyncService private constructor(
...
 
+    init {
```
LinX64, line 72: This is also being initialized right after the first constructor, shouldn't we put it at the top right after the curly bracket?

app/src/main/java/pl/llp/aircasting/ui/view/screens/login/LoginActivity.kt
```diff
@@ -17,6 +18,10 @@ class LoginActivity : BaseActivity() {
     @Inject
     lateinit var apiServiceFactory: ApiServiceFactory
 
+    @Inject
+    @Nullable
+    lateinit var sessionsSyncService: SessionsSyncService
```
LinX64, line 23: What's the point of initializing SessionsSyncService here and not using it? Am I missing something?

app/src/main/res/layout/logging_out_in_background.xml
xameYeoh: Alright, I will resist my urge to comply with Lint 😄

app/src/main/java/pl/llp/aircasting/data/api/services/SessionsSyncService.kt
xameYeoh: I think we can, but it is already used in quite a lot of places. They, in turn, will also need to be provided by dagger. I am refactoring as I go to the extent, but this change will trigger a lot of additional refactoring. I am not ready to do it in this PR

app/src/main/java/pl/llp/aircasting/ui/view/screens/login/LoginActivity.kt
xameYeoh: Thanks for mentioning! It is a leftover from some other approach I intended to use I guess

PR #642, opened 2022-08-26
Change lets begin background color to match rendering

xameYeoh:


LinX64 (changes requested):
Needs the dark color changing as well. (`values-night/colors.xml`) with the same ID.

xameYeoh:
> Needs the dark color changing as well. (`values-night/colors.xml`) with the same ID.
>
Change to what colour? This is the corresponding variant of the colour in the night styles:
`<color name="light_blue_lets_start_background">#40475D</color>`
Does it need to be different?

LinX64:
> > Needs the dark color changing as well. (`values-night/colors.xml`) with the same ID.
> 
> @LinX64 Change to what colour? This is the corresponding variant of the colour in the night styles: `<color name="light_blue_lets_start_background">#40475D</color>` Does it need to be different?

Absolutely, so we select the dark color based on the light tint, that is why when you change the light color, you will also need to select the corresponding light color for the Dark color for the UI/UX experience. 

Anyways, these colors should be provided by our designers or, on the Invision platform. You can also use this tool online: https://material.io/resources/color/ but still, we need to get the colors for both light/dark from the designers.

xameYeoh:
> > Absolutely, so we select the dark color based on the light tint, that is why when you change the light color, you will also
> > need to select the corresponding Dark color for a better UI/UX experience.
> 
> I know, the thing is that we already had a corresponding dark colour added, and it was okay - that's why I did not change anything there. Thanks for the rendering - I've picked a colour from there for the dark background as well. To me it seems similar to what we had there already, maybe just a tiny bit different 🤔
>
I somehow edited your comment instead of replying to it 😆

LinX64 (approved):

PR #643, opened 2022-08-26
mobile dormant & fixed: spinner obscured by map button

xameYeoh:


app/src/main/res/layout/session_card.xml
```diff
@@ -132,11 +132,13 @@
             android:id="@+id/loader"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:layout_marginTop="@dimen/keyline_8"
-            android:layout_marginBottom="@dimen/keyline_8"
+            android:layout_marginTop="@dimen/keyline_6"
```
LinX64, line 135: The top padding should be still reduced. I used `keyline_5` and it was looking much better.

LinX64 (changes requested):
Left comment

app/src/main/res/layout/session_card.xml
xameYeoh: I've placed it now using constraints, without margin. I think that if the margin is used there will still be devices where it will be obscured (on my phone for example I don't have it obscured even keyline_8 is used 😮 )

app/src/main/res/layout/session_card.xml
```diff
@@ -132,12 +132,14 @@
             android:id="@+id/loader"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:layout_marginTop="@dimen/keyline_8"
```
LinX64, line 135: Looks good with the new changes, however, this will cause an overlap which we will need to take care of it ASAP.

alt text

LinX64 (approved):

PR #644, opened 2022-08-26
Sessions list scroll on update crash

xameYeoh:
This replaces the hashmaps that were used around `SessionsObserver` with concurrent hashmaps to avoid ConcurrentModificationException


LinX64 (approved):
Looks good.

PR #646, opened 2022-08-31
Following > graph: time on right side of graph doesn't update

LinX64:


LinX64:
[following > graph: time on right side of graph doesn't update](https://trello.com/c/ubZ9dkjb/1635-following-graph-time-on-right-side-of-graph-doesnt-update)

xameYeoh (approved):

PR #647, opened 2022-08-31
Save wifi password

xameYeoh:
This utilises the Google autofill to save wifi credentials. It displays information bottomsheet in case google suspects the credentials to be weak

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/ui/view/screens/new_session/session_details/WeDoNotLeakPasswordsBottomSheet.kt
+++ b/app/src/main/java/pl/llp/aircasting/ui/view/screens/new_session/session_details/WeDoNotLeakPasswordsBottomSheet.kt
@@ -1,0 +1,18 @@
+package pl.llp.aircasting.ui.view.screens.new_session.session_details
+
+import android.text.SpannableStringBuilder
+import pl.llp.aircasting.R
+import pl.llp.aircasting.ui.view.screens.onboarding.your_privacy.LearnMoreYourPrivacyBottomSheet
+
+class WeDoNotLeakPasswordsBottomSheet : LearnMoreYourPrivacyBottomSheet() {
+    override val EXPANDED_PERCENT = 1.0
+
+    override fun layoutId(): Int {
+        return R.layout.we_do_not_leak_passwords
+    }
+
+    override fun buildDescription(): SpannableStringBuilder {
+        return SpannableStringBuilder()
+            .append(getString(R.string.we_do_not_leak_passwords))
+    }
+}
--- a/app/src/main/res/layout/network_password_dialog.xml
+++ b/app/src/main/res/layout/network_password_dialog.xml
@@ -1,46 +1,69 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.appcompat.widget.LinearLayoutCompat
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.appcompat.widget.LinearLayoutCompat xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:orientation="vertical"
-    android:background="@drawable/dialog_background">
+    android:background="@drawable/dialog_background"
+    android:orientation="vertical">
 
     <TextView
         android:id="@+id/wifi_password_dialog_header"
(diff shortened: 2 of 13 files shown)


app/src/main/java/pl/llp/aircasting/ui/view/screens/new_session/session_details/WeDoNotLeakPasswordsBottomSheet.kt
```diff
@@ -0,0 +1,18 @@
+package pl.llp.aircasting.ui.view.screens.new_session.session_details
+
+import android.text.SpannableStringBuilder
+import pl.llp.aircasting.R
+import pl.llp.aircasting.ui.view.screens.onboarding.your_privacy.LearnMoreYourPrivacyBottomSheet
+
+class WeDoNotLeakPasswordsBottomSheet : LearnMoreYourPrivacyBottomSheet() {
```
LinX64, line 7: Is there a specific reason for using the extended `LearnMoreYourPrivacyBottomSheet()` here? We have a base `BottomSheet()` class where we use it for all BottomSheets.

app/src/main/res/layout/we_do_not_leak_passwords.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
```
LinX64, line 1: When we create layouts, all of those(let's say most of) xml layouts should have a corresponding name like, `we_do_not_leak_passwords` -> `we_do_not_leak_passwords_bottom_sheet` based on the classes, so that it will be easier to find in large-scale projects and of course, more distinguishable in the res directory.

app/src/main/java/pl/llp/aircasting/ui/view/screens/new_session/session_details/WeDoNotLeakPasswordsBottomSheet.kt
xameYeoh: Thanks for mentioning that! The `LearnMoreYourPrivacyBottomSheet` had basically all the code that I needed for the `WeDoNotLeakPasswordsBottomSheet`, so I thought it would be less repetitive if I just extend it from there. Though then I made it a bit different with a separate layout, and now I can see that it'll be the same amount of code if we extend it from the Base class already, setting the text in xml directly. I will push a commit for that 🙏

app/src/main/res/layout/we_do_not_leak_passwords.xml
xameYeoh: Thanks for mentioning, I've included that in the last commit

app/src/main/res/layout/network_password_dialog.xml
```diff
@@ -1,55 +1,90 @@
...
+        style="@style/Widget.Aircasting.TextInputLayout"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
         android:layout_marginStart="@dimen/keyline_8"
+        android:layout_marginTop="@dimen/keyline_6"
         android:layout_marginEnd="@dimen/keyline_8"
-        android:layout_marginBottom="@dimen/keyline_6" />
+        app:hintEnabled="false">
+
+        <com.google.android.material.textfield.TextInputEditText
```
LinX64, line 43: We already have a style for `TextInputLayout`s right? : `style="@style/Widget.Aircasting.TextInputEditTextFilledBox"`

app/src/main/res/layout/we_do_not_leak_passwords_bottom_sheet.xml
```diff
@@ -0,0 +1,82 @@
...
+                android:paddingStart="@dimen/keyline_4"
+                android:src="@drawable/ic_x"
+                app:layout_constraintBottom_toTopOf="@+id/learn_more_onboarding_your_privacy_header"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintTop_toTopOf="parent"
+                app:tint="@color/aircasting_blue_400" />
+
+            <TextView
+                android:id="@+id/learn_more_onboarding_your_privacy_header"
+                style="@style/TextAppearance.Aircasting.NewSessionDetails.Headline1"
```
LinX64, line 36: We have a style for font in here, and also in here: `android:textAppearance="@style/TextAppearance.Aircasting.Headline2"`

Shouldn't it be only **one** style with customized text?

LinX64 (approved):
Looks good after those changes.

PR #662, opened 2022-09-06
Graph Zoom | EndTime text

LinX64:


LinX64:
[following > graph: time on right side of graph doesn't update](https://trello.com/c/ubZ9dkjb/1635-following-graph-time-on-right-side-of-graph-doesnt-update)

LinX64:
[Following > third party sessions > graph: midnight line not accurate](https://trello.com/c/roVpz9TK/1670-following-third-party-sessions-graph-midnight-line-not-accurate)

LinX64:
This fixes the problem with updating the graph, but it will cause another jumping issue with zooming out!

PR #665, opened 2022-09-07
Midnight line external airbeam wrong time 

xameYeoh:


app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -0,0 +1,42 @@
...
+    set(HOUR_OF_DAY, 0)
+    set(MINUTE, 0)
+    set(SECOND, 0)
+    set(MILLISECOND, 0)
+
+    return timeInMillis / 1000
+}
+
+fun Calendar.getEndOfTodaySeconds(): Long {
+    // UTC timezone here is ESSENTIAL, as backend assumes receiving timestamp in UTC
```
LinX64, line 34: It might be clear to us, but it won't be for the new developers who might work on it. Besides it seems strange to have comments larger in size than actual methods

app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -0,0 +1,42 @@
...
+    add(YEAR, -1)
+    set(HOUR_OF_DAY, 0)
+    set(MINUTE, 0)
+    set(SECOND, 0)
+    set(MILLISECOND, 0)
+
+    return timeInMillis / 1000
+}
+
+fun Calendar.getEndOfTodaySeconds(): Long {
```
LinX64, line 33: "Besides it seems strange to have comments larger in size than actual methods"

It might be clear to us, but again, think of future where there might be new developers working on this. We should either document this or, add it here and AFAIR, we haven't decided to document these because of the time limitation, so I would keep the description here for now.

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
```diff
@@ -0,0 +1,25 @@
+package pl.llp.aircasting.util
+
+import java.util.*
+
+class DateUtil {
```
LinX64, line 5: And here, we're back to our discussion: https://github.com/HabitatMap/AircastingAndroid/pull/636#discussion_r955772518

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/graph/GraphContainer.kt
```diff
@@ -367,12 +369,11 @@ class GraphContainer(
 
     private fun updateVisibleTimeSpan() {
         mGraph?.let { graph ->
+            // sets to almost epoch???
```
LinX64, line 372: If we're not sure what that is, maybe we shouldn't add a comment in there? what do you think?

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/graph/GraphDataGenerator.kt
```diff
@@ -17,9 +18,9 @@ class GraphDataGenerator(
     private var cumulativeValue = 0.0
     private var cumulativeTime: Long = 0
     private var count = 0
+    private var startTime = Date()
```
LinX64, line 21: We're initialising it in this line, right? `startTime = firstMeasurement.time` - the variable is suppose to hold a date, which we're initialising it later in the class, so maybe you forgot to make it `lateinit` as I did before? : `private lateinit var startTime: Date`

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/graph/GraphContainer.kt
```diff
@@ -346,15 +347,16 @@ class GraphContainer(
         updateVisibleTimeSpan()
     }
 
+    @OptIn(DelicateCoroutinesApi::class)
     private fun updateLabelsBasedOnVisibleTimeSpan() {
         mGraph ?: return
 
-        runOnIOThread {
+        GlobalScope.launch(Dispatchers.IO) {
```
LinX64, line 354: Shouldn't we use the refactored extension func? `runOnIOThread{}` - there is no point in keeping the previous `GlobalScope()`.

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/graph/GraphContainer.kt
xameYeoh: Leftover, sorry 😄

app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
xameYeoh: Okay, but I think we should put something different in the comments there - as Why we are using the method, not What it does. Because you can see from the code What it does. And frankly, reading it confuses me even more

xameYeoh: How about this text:
The method is used to set left time boundary for search of sessions on google map in S&F feature Setting UTC time zone is essential, as all time communicated to backend is assumed to be in UTC Example:
08/09/2022 10:30:00 GMT+2 - Current local timestamp
08/09/2021 00:00:00 GMT+0 - Result - start of current day a year ago in UTC
Returns:
1631059200
`fun Calendar.getStartOfTodayEpochFromYearAgo()`

app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -0,0 +1,58 @@
...
+}
+
+fun Calendar.dayOfMonth(date: Date, isExternalSession: Boolean): Int {
+    time = date
+    if (isExternalSession)
+        timeZone = TimeZone.getTimeZone("UTC")
+    return this[DAY_OF_MONTH]
+}
+/**
+ * The method is used to set left time boundary for search of sessions on google map in S&F feature
```
LinX64, line 21: It's now an extension function(`Calendar.getStartOfTodayEpochFromYearAgo`), rather than a method.

How about this: **This extension function is used to set the left time boundary for searching of sessions on Google Map in S&F feature.**

app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Setting UTC time zone is essential, as all time communicated to backend is assumed to be in UTC
```
LinX64, line 22: Maybe this would be a better explanation? **Setting the UTC time zone is essential, as all-time communicated to the backend is assumed to be in the UTC time zone.**

LinX64 (approved):
I left some comments for the explanations, overall looks good.

LinX64:
[BottomSheet Styles](https://trello.com/c/EhGGJkBB/1772-bottomsheet-styles)

PR #666, opened 2022-09-07
[2] Midnight line on external sensors

xameYeoh:
This is a 2nd part of issue with midnight line - making it work with 3rd party sensors like OpenAQ

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
+++ b/app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
@@ -1,5 +1,6 @@
 package pl.llp.aircasting.util.extensions
 
+import org.apache.commons.lang3.time.DateUtils
 import java.util.*
 import java.util.Calendar.*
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -13,11 +14,22 @@ fun Calendar.addHours(date: Date, hours: Int): Date {
 
 fun Calendar.dayOfMonth(date: Date, isExternalSession: Boolean): Int {
     time = date
+    // UTC timezone here is ESSENTIAL, as external sessions timestamps come in UTC from backend
```
LinX64, line 17: Duplicate comment; Do we really need that?

app/src/main/java/pl/llp/aircasting/util/extensions/CalendarExtensions.kt
```diff
@@ -13,11 +14,22 @@ fun Calendar.addHours(date: Date, hours: Int): Date {
...
+    time = date
+    // UTC timezone here is ESSENTIAL, as external sessions timestamps come in UTC from backend
+    if (isExternalSession)
+        timeZone = TimeZone.getTimeZone("UTC")
+
+    return truncate(this, DAY_OF_MONTH).time
+}
+
 fun Calendar.getStartOfTodaySecondsFromYearAgo(): Long {
     // UTC timezone here is ESSENTIAL, as backend assumes receiving timestamp in UTC
```
LinX64, line 34: Instead of having almost the same comments for each extension, you could easily add a multiline comment as a general comment at the top of those extensions to make it more clear and adding specific comment for each extension.

It doesn't make sense to have 4 similar comments in each extension.

app/src/main/java/pl/llp/aircasting/util/DateUtil.kt
```diff
@@ -0,0 +1,25 @@
+package pl.llp.aircasting.util
+
+import java.util.*
+
+class DateUtil {
```
LinX64, line 5: This is not being used. Maybe you forgot to remove it?

PR #671, opened 2022-09-09
Update Graph end time

xameYeoh:
zoomToDefault and update labels only when Graph is fully zoomed out


LinX64 (approved):
Looks good! Nice catch.

PR #675, opened 2022-09-12
Attach a note | Show note image | Load from DB

LinX64:
This is the first implementation for showing the note image on the note.


LinX64:
[Attach a Note: Showing Notes Images](https://trello.com/c/hVFdu2at/1766-attach-a-note-showing-notes-images)

app/src/main/res/layout/add_note_bottom_sheet.xml
```diff
@@ -110,11 +110,11 @@
                     <com.google.android.material.textfield.TextInputEditText
                         android:id="@+id/note_input"
                         android:layout_width="match_parent"
-                        android:layout_height="wrap_content"
+                        android:layout_height="150dp"
```
xameYeoh, line 113: I am not sure about the context, is there no way to make it non-fixed size here? If so, I would add some dimen resource, or maybe there is already one with similar size for the purpose

app/src/main/res/layout/edit_note_bottom_sheet.xml
```diff
@@ -37,140 +40,152 @@
...
                     android:layout_height="wrap_content"
-                    android:text="You can edit your note here"
-                    android:textAppearance="@style/TextAppearance.Aircasting.Body1"
                     android:layout_marginStart="@dimen/keyline_6"
+                    android:layout_marginTop="@dimen/keyline_2"
                     android:layout_marginEnd="@dimen/keyline_6"
-                    android:layout_marginTop="@dimen/keyline_6"
-                    app:layout_constraintTop_toBottomOf="@id/title"
-                    app:layout_constraintBottom_toTopOf="@id/note_description"
+                    android:text="You can edit your note here"
```
xameYeoh, line 85: I guess we could add string resource for that

xameYeoh (approved):

app/src/main/res/layout/edit_note_bottom_sheet.xml
LinX64: Surprisingly, we didn't have translations for the edit note part.

PR #678, opened 2022-09-13
Sessions not being synced after Attach note

LinX64:


LinX64:
[Sessions not being synced after Attach note](https://trello.com/c/R3MUY7Mb/1774-sessions-not-being-synced-after-attach-note)

LinX64:
A quick note: For better performance, I used `Dispatchers.Main`: [Best practice](https://developer.android.com/codelabs/kotlin-coroutines#8)

**_Both Room and Retrofit make suspending functions main-safe.

It's safe to call these suspend funs from Dispatchers.Main, even though they fetch from the network and write to the database._**

xameYeoh (approved):

PR #684, opened 2022-09-14
Show image note after logout

xameYeoh:
This gets the image path from server and appends backend url to it to download image with Glide to the bottomsheet

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/util/Settings.kt
+++ b/app/src/main/java/pl/llp/aircasting/util/Settings.kt
@@ -1,6 +1,7 @@
 package pl.llp.aircasting.util
 
 import android.content.SharedPreferences
+import pl.llp.aircasting.data.api.util.ApiConstants
 
 open class Settings(private val sharedPreferences: SharedPreferences) {
     companion object {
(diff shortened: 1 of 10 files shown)


app/src/main/java/pl/llp/aircasting/util/Settings.kt
```diff
@@ -46,8 +47,6 @@ open class Settings(private val sharedPreferences: SharedPreferences) {
         private const val DEFAULT_ONBOARDING_DISPLAYED = false
     }
 
-    protected open val DEFAULT_BACKEND_URL = "http://aircasting.org"
```
LinX64, line 49: We created this companion object for holding only settings stuff.

We also have the same url in the ApiConstants and since now it is not private, how about removing this one from here and using the one we have in ApiConstants?

LinX64 (approved):
I left a minor comment, overall it looks good.

app/src/main/java/pl/llp/aircasting/util/Settings.kt
```diff
@@ -38,6 +38,7 @@ open class Settings(private val sharedPreferences: SharedPreferences) {
         private const val DEFAULT_MAPS_DISABLED = false
 
         private const val DEFAULT_BACKEND_PORT = "80"
+        const val DEFAULT_BACKEND_URL = "http://aircasting.org"
```
LinX64, line 41: Single responsibility : this is one of the SOLID principles. By having only one constant and baseUrl we will be able to follow the SOLID principles and this will allow us to keep Settings from Api Service separated.

app/src/main/java/pl/llp/aircasting/util/Settings.kt
xameYeoh: True, completely forgot about that

PR #697, opened 2023-09-08
Feat/add location for abm sync

Sebenicz:


app/src/main/java/pl/llp/aircasting/util/helpers/sensor/airbeam3/sync/sessionProcessor/SDCardMobileSessionsProcessor.kt
```diff
@@ -28,20 +28,28 @@ class SDCardMobileSessionsProcessor @AssistedInject constructor(
...
             Log.v(TAG, "Could not find session with uuid: ${csvSession.uuid} in DB")
             session = csvSession.toSession(deviceId, lineParameterHandler.deviceType) ?: return
             sessionId = mSessionsRepository.insert(session)
         } else {
             session = Session(dbSession)
             sessionId = dbSession.id
+            val measurementStreamId = mMeasurementStreamsRepository.getStreamsIdsBySessionId(sessionId).last()
+            val measurementsInDB = mMeasurementsRepository.getBySessionIdAndStreamId(sessionId, measurementStreamId)
+            lastLat = measurementsInDB.findLast { it?.latitude != null }?.latitude
+            lastLng = measurementsInDB.findLast { it?.longitude != null }?.longitude
```
xameYeoh, line 52: I think there is a lastMeasurementTime(sessionId: Long?) method in MeasurementsRepository that you could utilise directly. While it does return a nullable value, if there is an active session in the DB prior to the sync, then we should have a last measurement with location there (I am pretty sure the active measurements always get the location nervous-laugh).

If we still want to make sure we find the measurement with the non-null coordinates, I would create a new lastMeasurementTime() method and a new Room sql query for that (with the same logic you did in the code). So that the return value is just a location (either two method for lng and lat, or maybe we can have a Pair<Double,Double> there? Not sure if Room supports it out of the box as the return type)

No worries if the query and method will only be used for this specific case - it's quite normal, they don't have to be universal. 

The thing with AirCasting app is that it operates on large amounts of data (some people have lots and lots of measurements in a session), so if we need just a couple of values from the DB, it's always better to get them right on the query level, instead of fetching all the measurements and then searching through them, as it can compromise performance and even lead to some exceptions (we've got some live data crashes on firebase one of the reasons for which could be an attempt to fetch a lot of data in one time)

xameYeoh: So, this is the rare project where you need to think about performance a bit 😄

xameYeoh (approved):

PR #699, opened 2023-09-13
Show AB battery level

xameYeoh:


app/src/main/java/pl/llp/aircasting/util/helpers/sensor/airbeamSyncable/AirBeam3Reader.kt
```diff
@@ -22,7 +23,17 @@ class AirBeam3Reader(
...
 
         if (dataString.isNotEmpty()) {
-            val newMeasurementEvent = responseParser.parse(dataString)
+
+            val regex = """([\d.]+)V\s+(\d+)%""".toRegex()
+            val matchResult = regex.find(dataString)
+
+            val newMeasurementEvent = if (matchResult != null){
+                val (volts, percentage) = matchResult.destructured
+                NewBatteryReadingEvent(volts.toDouble(), percentage.toInt())
```
xameYeoh, line 32: "You were meant to destroy the sith not join them!" :D 
I wonder if we could achieve communication of data to the service via state/shared flow, and not continuing using the EventBus (I know we talked about consistency, but this one is of the rustiest parts that we talked about we wanted to change, and it's worth to slowly get rid of it, or at least not use it further)

xameYeoh: And I would also add private methods with names here, so that the main flow would be readable, something like:
if (data string is battery level measurement) 
  updateCurrentBatteryLevel()
else
  postNewMeasurementFromSensorEvent()

xameYeoh (commented):
Great job! I've left some comments to consider. I think we might try to use flows here instead of EventBus. Also, please don't forget to format the code (it should become automatic - write a few lines, smash the option+cmd+L, write a few lines, smash the option+cmd+L...)

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/airbeamSyncable/AirBeam3Reader.kt
```diff
@@ -13,18 +16,42 @@ import javax.inject.Inject
...
+        } else null
+    }
+
+    private fun updateBatteryLevel(percentage: Int) {
+        coroutineScope.launch {
+            batteryLevelFlow.emit(percentage)
+        }
+    }
+
+    private fun postNewMeasurementEvent(measurement: String) {
```
xameYeoh, line 54: I like the new method structure, looks clean. Good job!

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/services/BatteryLevelService.kt
```diff
@@ -0,0 +1,139 @@
...
+
+            notificationManager.notify(
+                BATTERY_LEVEL_NOTIFICATION_ID,
+                createNotification("$batteryLevel%")
+            )
+            batteryLastState = batteryLevel
+        }
+    }
+
+    private fun isOutOfHysteresis(batteryLevel: Int): Boolean {
```
xameYeoh, line 99: Wow, now that's the word 😄

xameYeoh: I usually get hysteresis when I see Java

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/services/BatteryLevelService.kt
```diff
@@ -0,0 +1,139 @@
...
+        val upperBoundary = batteryLastState + 2
+        val lowerBoundary = batteryLastState - 3
+        return batteryLevel !in lowerBoundary..upperBoundary
+    }
```
xameYeoh, line 103: That's a clean implementation! I like the readability

PR #704, opened 2023-11-30
Fix/stuck at connecting

Sebenicz:
When connecting to AB in NewSessionActivity without granted permissions, app gets stuck on ConnectingAirbeamFragment because event posted when connection Is established (that navigates to next fragment) is not received by NewSessionController. Changing event to a flow fixed this issue. 

I've looked at the possibility of changing the rest of the connection logic that is currently working on EventBus and it seems like it's going to take too much time.

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/data/model/AirbeamConnectionStatus.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/model/AirbeamConnectionStatus.kt
@@ -1,0 +1,9 @@
+package pl.llp.aircasting.data.model
+
+import pl.llp.aircasting.ui.view.screens.new_session.select_device.DeviceItem
+
+data class AirbeamConnectionStatus(
+    val deviceItem: DeviceItem? = null,
+    val sessionUUID: String? = null,
+    val isConnected: Boolean,
+)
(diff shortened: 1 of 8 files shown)


app/src/main/java/pl/llp/aircasting/data/model/AirbeamConnectionStatus.kt
```diff
@@ -0,0 +1,9 @@
+package pl.llp.aircasting.data.model
```
Sebenicz, line 1: I had no clue where to place this class, So I'm open to any suggestions here 😅

xameYeoh: I guess it's a model, so we are good :)

app/src/main/java/pl/llp/aircasting/di/modules/FlowModule.kt
```diff
@@ -12,8 +14,17 @@ class FlowModule {
...
 
 @Retention(AnnotationRetention.BINARY)
 @Qualifier
-annotation class BatteryLevelFlow
\ No newline at end of file
+annotation class BatteryLevelFlow
+
+@Retention(AnnotationRetention.BINARY)
+@Qualifier
+annotation class AirbeamConnectionStatusFlow
```
xameYeoh, line 30: Mua! Chef's kiss

xameYeoh (approved):
Great stuff! Awesome to know we now have one more flow and one less event 💪

PR #705, opened 2023-12-28
Fix/mobile sync

Sebenicz:
Added flow with boolean to stop reconnection service when user starts sync while AB is in the standalone mode which could cause session to go into "recording" state after sync.

after ABM restarts during fixed session, we receive fake location, to not display it only the location with first coordinates is used

### DIFF
--- a/app/src/main/java/pl/llp/aircasting/data/model/Session.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/model/Session.kt
@@ -2,6 +2,7 @@ package pl.llp.aircasting.data.model
 
 import pl.llp.aircasting.data.api.response.search.SessionInRegionResponse
 import pl.llp.aircasting.data.local.entity.CompleteSessionDBObject
+import pl.llp.aircasting.data.local.entity.LatLng
 import pl.llp.aircasting.data.local.entity.SessionDBObject
 import pl.llp.aircasting.data.local.entity.SessionWithNotesDBObject
 import pl.llp.aircasting.data.local.entity.SessionWithStreamsAndLastMeasurementsDBObject
@@ -214,9 +215,11 @@ open class Session(
         }
 
         override fun equals(other: Any?): Boolean {
-            if (other == null || other !is Location) return false
-
-            return latitude == other.latitude && longitude == other.longitude
+            return when (other){
+                is Location -> latitude == other.latitude && longitude == other.longitude
+                is LatLng -> latitude == other.latitude && longitude == other.longitude
+                else -> false
+            }
         }
 
         override fun hashCode(): Int {
--- a/app/src/main/java/pl/llp/aircasting/data/model/Session.kt
+++ b/app/src/main/java/pl/llp/aircasting/data/model/Session.kt
@@ -215,7 +215,7 @@ open class Session(
         }
 
         override fun equals(other: Any?): Boolean {
-            return when (other){
+            return when (other) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/pl/llp/aircasting/data/model/Session.kt
```diff
@@ -214,9 +215,11 @@ open class Session(
...
 
         override fun equals(other: Any?): Boolean {
-            if (other == null || other !is Location) return false
-
-            return latitude == other.latitude && longitude == other.longitude
+            return when (other) {
+                is Location -> latitude == other.latitude && longitude == other.longitude
+                is LatLng -> latitude == other.latitude && longitude == other.longitude
+                else -> false
+            }
```
xameYeoh, line 222: I wouldn't use equals() to compare one type to another, to me it seems like it's outside of its "contract" and non standard. 

Because if we are saying that Location.equals(LatLng) can be true, then LatLng.equals(Location) should also be possible, but it can't, since we are not in charge of the LatLng class - it's provided by the library. But even if we could, I still think we shouldn't

Instead I would define a short conversion function like that:
fun Session.Location.toLatLng() = LatLng(latitude, longitude) 
And use equals() with one type

app/src/main/java/pl/llp/aircasting/ui/view/screens/session_view/map/MapContainer.kt
```diff
@@ -170,17 +170,28 @@ class MapContainer(rootView: View?, context: Context, supportFragmentManager: Fr
...
                 measurement,
                 mSessionPresenter?.selectedSensorThreshold()
             )
-            latestPoint = LatLng(measurement.latitude!!, measurement.longitude!!)
-            mMeasurementPoints.add(latestPoint)
-            i += 1
+            if (!isFixed) {
+                val measurementLocation = LatLng(measurement.latitude!!, measurement.longitude!!)
+
+                if (!Session.Location.FAKE_LOCATION.equals(measurementLocation)) {
```
xameYeoh, line 189: Here, if we are going with the conversion function we could use the == and != respectively:

val measurementLocation = Session.Location(measurement.latitude!!, measurement.longitude!!)

                if (Session.Location.FAKE_LOCATION != measurementLocation) {
                    latestPoint = measurementLocation.toLatLng()
                }

app/src/main/java/pl/llp/aircasting/util/helpers/sensor/common/connector/AirBeamReconnector.kt
```diff
@@ -26,6 +28,8 @@ class AirBeamReconnector(
     private val coroutineScope: CoroutineScope,
     private val sessionUuidByStandaloneMode: MutableMap<String, Boolean> = ConcurrentHashMap(),
     private val connectionStatusFlow: MutableStateFlow<AirbeamConnectionStatus>,
+    @SyncActiveFlow
+    private val syncStatusFlow: MutableSharedFlow<Boolean>,
```
xameYeoh, line 32: I think we should find a way to provide a non-mutable shared flow to the observers, so that there will be no way of emitting to the flow from anywhere outside the class responsible for the Sync

xameYeoh (approved):
Yay, new flow!
I've left some comments for consideratio

xameYeoh (approved):
Cool!

PR #707, opened 2024-04-03
feat:Delete account confirmation

mpanuszewska:
This should not be merged until the backend is deployed and the adress is changed. 

UI: 
Screenshot 2024-04-05 at 13 32 42
Screenshot 2024-04-05 at 14 30 56


app/src/main/java/pl/llp/aircasting/data/api/params/DeleteAccountConfirmationCodeBody.kt
```diff
@@ -0,0 +1,4 @@
+package pl.llp.aircasting.data.api.params
+
+class DeleteAccountConfirmationCodeBody(val code: String) {
+}
```
xameYeoh, line 4: This could be done as a data class (aka model class, we use them for plain data representation and they have a couple of things out of the box that are useful compared to default classes - https://kotlinlang.org/docs/data-classes.html):

data class DeleteAccountConfirmationCodeBody(val code: String)

app/src/main/java/pl/llp/aircasting/data/api/response/DeleteAccountSendEmailResponse.kt
```diff
@@ -0,0 +1,3 @@
+package pl.llp.aircasting.data.api.response
+
+class DeleteAccountSendEmailResponse
```
xameYeoh, line 3: This could be a data class too

app/src/main/java/pl/llp/aircasting/data/api/util/ApiConstants.kt
```diff
@@ -1,7 +1,7 @@
 package pl.llp.aircasting.data.api.util
 
 object ApiConstants {
-    const val baseUrl = "http://aircasting.org"
+    const val baseUrl = "http://23.92.17.56"
```
Sebenicz, line 4: Looks like a test URL sneaked in 😅 🥷

app/src/main/res/layout/confirm_code_dialog.xml
```diff
@@ -0,0 +1,58 @@
...
+        android:layout_height="wrap_content"
+        style="@style/TextAppearance.Aircasting.NewSessionDetails.Headline1"
+        android:gravity="center"
+        android:textSize="@dimen/text_size_m"
+        app:lineHeight="@dimen/keyline_6"
+        tools:text="@string/my_account_delete_confirmation_code"
+        android:layout_marginTop="@dimen/keyline_8"
+        android:layout_marginStart="@dimen/keyline_8"
+        android:layout_marginEnd="@dimen/keyline_8"
+        android:layout_marginBottom="@dimen/keyline_6" />
```
Sebenicz, line 22: Just a suggestion: the whole dialog text is in "header"/"title" style, maybe we could change it so it is displayed as a body and do a title like "Confirm delete" or something like that 😅   dialog_alert_aircasting.xml could be used as a reference here for text styles and sizes.

app/src/main/java/pl/llp/aircasting/data/api/util/ApiConstants.kt
mpanuszewska: Hence the "This should not be merged until the backend is deployed and the adress is changed." comment but good to know you're looking too :D

xameYeoh (commented):
Great job! I've left a few comments to consider

app/src/main/java/pl/llp/aircasting/data/api/response/DeleteAccountSendEmailResponse.kt
mpanuszewska: This one is empty so technically it can't, since "The primary constructor must have at least one parameter.", right? :)

PR #717, opened 2024-10-24
fix: change AB icon background to transparent

mpanuszewska:


Sebenicz (approved):
now that's optimization <3

PR #721, opened 2024-12-03
feat: Hide data transmission toggle for fixed sessions for ABMs

mpanuszewska:


Sebenicz (approved):
very nice change, very readable, very demure, LLKTM

